PR #526, opened 2023-07-12
added the igbo and javanese lang

rakshith-arch:


qa_scripts/missing_foilstones_paths.txt
Rajesh1041, line 1: You want to commit this file too?

PR #990, opened 2024-02-27
updated tapped_start and selected_level events 

Nikhil77sutara:
• Updated tapped_start and selected_levels as custom events.
Passing all the required params in the events.
•	Removing event_date_with_timestamp.


src/components/play-button.ts
```diff
@@ -29,6 +35,13 @@ export default class PlayButton {
             this.imagesLoaded = true;
         });
     }
+    private async init() {
```
Rajesh1041, line 38: Can we make correct spacing!

src/scenes/level-selection-scene.ts
```diff
@@ -35,7 +40,9 @@ export class LevelSelectionScreen {
         ? Math.floor(self.data.levels.length / 10) + 1
         : Math.floor(self.data.levels.length / 10);
     this.initialiseButtonPos();
-    this.levels = [];
+    this.levels = [];  
+    this.firebaseIntegration = new FirebaseIntegration();
+     this.init();
```
Rajesh1041, line 45: Please do correct formatting here!

src/scenes/level-selection-scene.ts
```diff
@@ -71,6 +78,13 @@ export class LevelSelectionScreen {
     });
     this.addListeners();
   }
+  private async init() {
+
```
Rajesh1041, line 82: Can we do correct spacing here!

src/scenes/start-scene.ts
```diff
@@ -12,7 +12,7 @@ import { AudioPlayer } from "../components/audio-player";
 import { FirebaseIntegration } from "../Firebase/firebase-integration";
 import { Utils } from "../common/utils";
 import PlayButton from "../components/play-button";
-
+import { TappedStart } from "../Firebase/firebase-event-interface";
```
Rajesh1041, line 15: Why we are importing `TappedStart `if we are not using it?

Rajesh1041:
Can we add one more point to description?
- Removing event_date_with_time

Rajesh1041 (approved):

PR #1028, opened 2024-03-21
added download_completed firebase event

ashishkrr:
logging the download_completed firebase event which fires right after the service worker completes 100% cache is done.


Rajesh1041 (approved):

Rajesh1041 (commented):
Can we include some useful description?

Rajesh1041 (approved):

Issue #1027, opened 2024-03-20
Allow user to tap anywhere on splash screen to move onto the level selection screen


burrage:
**User Story**
As a UX designer and Product Owner,
I want to allow the user to tap anywhere on the FTM JS splash screen (except the "X" to close the app and return to the app selection screen!) to move to the level selection page,
So that we move the user on to level selection faster and we update the event to how we are logging our "tapped_start" event (it's logging if someone taps anywhere on that screen).

**Acceptance Criteria**
Given that I am on the FTM JS splash screen,
When I tap anywhere but the "X" close button,
Then I should be taken to the level selection screen and a "tapped_start" event should have been logged.

Issue #1021, opened 2024-03-18
Randomly play 1 of 3 positive monster SFX after the "eat" SFX when getting correct answer


burrage:
**User Story**
As a UX designer,
I want to randomly play 1 of 3 monster cheering sounds after the "eating" sound when someone gets a correct answer,
So we are increasing the amount of positive SFX played when getting an answer correct to keep the child engaged in the task we want them to do (keep getting correct answers!)

Cheering 1: https://github.com/curiouslearning/FeedTheMonster/blob/master/Feed%20the%20Monster/Assets/Sounds/Monsters/Cheering-01.mp3
Cheering 2: https://github.com/curiouslearning/FeedTheMonster/blob/master/Feed%20the%20Monster/Assets/Sounds/Monsters/Cheering-02.mp3
Cheering 3:
https://github.com/curiouslearning/FeedTheMonster/blob/master/Feed%20the%20Monster/Assets/Sounds/Monsters/Cheering-03.mp3

**Acceptance Criteria**
Given that I am playing Feed The Monster,
When I get an answer correct,
Then I should hear the eating SFX followed by 1 of the 3 cheering SFX that are chosen at random.

PR #1031, opened 2024-03-26
added sfx eat audio and Allow user to tap anywhere on splash screen

ashishkrr:
added sfx eat audio which plays randomly among those, when we drop right answer stone #1021

Allow user to tap anywhere on splash screen to move onto the level selection screen and excluded the close button area for tapping #1027


Rajesh1041:
I can see that two audios are added, modify the title to reflect the same.
and please try to create separate PR according to different task.

ashishkrr:
> I can see that two audios are added, modify the title to reflect the same. and please try to create separate PR according to different task.


One is already added rajesh, so that is reason i just have added rest two which are not present before.

Rajesh1041:
I know that one audio is already present, i can see that too but title seems confusing. Also create separate PR for separate tickets since these tasks are not even related.

Rajesh1041:
@burrage Hey Ben this PR has two tasks in same PR, i feel its cleaner to have each task have their separate PRs. Should we follow that and create separate PR for this one or should i merge it.

burrage:
@Rajesh1041 @ashishkrr In general, we should stick to the 1 task = 1 PR model because it just makes life easier on the reviewer, merger, production maintainer, rollbacker...

Are the only changes for #1027 in start-scene.js? Since it is 3 lines and these are both pretty minor changes, we can make a one-time exception but in the future, let's please try to stick to 1 task = 1 PR. Thanks!

ashishkrr:
> @Rajesh1041 @ashishkrr In general, we should stick to the 1 task = 1 PR model because it just makes life easier on the reviewer, merger, production maintainer, rollbacker...
> 
> Are the only changes for #1027 in start-scene.js? Since it is 3 lines and these are both pretty minor changes, we can make a one-time exception (only if Rajesh feels it is appropriate) but in the future, let's please try to stick to 1 task = 1 PR as a best practice. Thanks!

@burrage Thanks for clarification, will follow the 1 task = 1 PR, @Rajesh1041 I am closing this PR as you asked me earlier to create separate PR for good practice.

PR #1046, opened 2024-03-27
Tapping hand tutorial to make stones appear faster.

rakshith-arch:
1. The PR is for tapping animation.
2. It appears in first screen when user starts to play.
3. The animation will disappear once the monster is clicked.


src/components/tutorial.ts
```diff
@@ -85,6 +89,29 @@ export class Tutorial {
...
+        const offsetY = this.height / 1.9 + (this.tutorialImg.height / 2);
+        this.context.drawImage(this.tutorialImg, offsetX, offsetY, scaledWidth, scaledHeight);
+    }
+    
+    sinusoidalInterpolation(time, minScale, maxScale, duration) {
+        const amplitude = (maxScale - minScale) / 2;
+        const frequency = Math.PI / duration;
+        return minScale + amplitude * Math.sin(frequency * time);
+    }
+
```
Rajesh1041, line 110: Can we remove unnecessary line spaces?

rakshith-arch: Hey @Rajesh1041 i have added that line spaces for the readability, where first 2 is for declaration, next one is for function call and the last one is for the logic, purposefully i added it, should i remove that?
and one more edge case was missing here i am working on it, i will fix that and and then let you know..

Rajesh1041: @rakshith-arch then spaces should be fine! Let me know when you are done with changes.

Issue #1004, opened 2024-03-06
Repeat dragging stone tutorial animation


burrage:
**User Story**
As a UX designer,
I want to repeat the stone-dragging tutorial animation multiple times on the first question of the first level,
So that if a child misses the animation the first time or they do not understand what to do, the animation of what they are supposed to do repeats multiple times until the timer runs out.

[FTM Unity tutorial reference video](https://drive.google.com/file/d/1IKr9t2IjyttKlBbygi13_AO7H-QJ12iQ/view?usp=sharing)

**Acceptance Criteria**
Given that I am playing the first question of the first level,
When the stones appear,
I want the stone-dragging animation to play.

Given that the stone-dragging animation has played in its entirety,
When an answer has not been given,
I want to repeat the stone-dragging tutorial animation on a loop until the child 1.) drags and releases a stone on the monster or 2.) the timer runs out.

PR #1060, opened 2024-03-29
#1004 Added repeatation of tutorial animation!

amitsinghsutara:
In this pr I have added the functionality of repeating. the tutorial animation again and again until the timer runs out or the user have finished giving input!


src/components/tutorial.ts
Rajesh1041, line 1: Can we also add stopping condition when user tap on monster?

amitsinghsutara: Rajesh the pr is good to go since we didn't have the functionality of stopping the tutorial animation when the user drags the stone in the unity version also!

Rajesh1041: okay then!

Rajesh1041 (approved):

PR #1108, opened 2024-04-12
updated fantastic audio

Nikhil77sutara:
- Added fantastic audio to the feedback audio list for us english.


src/components/stone-handler.ts
```diff
@@ -302,6 +303,7 @@ export default class StoneHandler extends EventManager {
       Utils.getConvertedDevProdURL(this.feedbackAudios[feedBackIndex]),
     );
     // to play the audio parrallely.
+    console.log("this is util fun>>>>"+Utils.getConvertedDevProdURL(this.feedbackAudios[feedBackIndex]))
```
Rajesh1041, line 306: Can we remove this console statement?

PR #1163, opened 2024-05-19
fix: fm-15 Added the ability to draw feedback texts in second line!

amitsinghsutara:
I have added the ability to draw the feedback texts in the second line, If the text is too long!


src/components/feedback-particle-effect/feedback-text-effects.ts
```diff
@@ -83,11 +83,25 @@ export class FeedbackTextEffects {
...
+        let lastText=" "+text.slice(text.lastIndexOf(" ",text.lastIndexOf(" ")-1)+1);
+        this.context.fillText(
+          initialText,
+          this.textX,
+          this.textY + index * this.lineHeight 
+        );
+        this.context.fillText(
+          lastText,
+          this.textX,
+          this.textY + index * this.lineHeight +50
```
Rajesh1041, line 97: Should try to ignore hardcoded values!

Rajesh1041 (approved):

PR #1193, opened 2024-07-24
fixed: [optimization] Fixed issue on word puzzle where feedback audio plays every correct letter instead of completion of the puzzle.

bernhardccodev:
Fixed by adding the original condition for checking if the whole puzzle is completed before playing the feedback audio.


ashwinnair-chimple (approved):

burrage:
@bernhardccodev @ashwinnair-chimple Hey guys! A heads up, we decided to change this from the Unity version to help reinforce to the kids that they were on the right path after each stone they dragged so this was not a bug.

Also differing from the Unity version, we reset the timer after each correct stone dragged as we found children with dexterity issues had a harder time when they made it to the spelling levels. There are also some languages where words can be up to 6+ characters so even the really fast kids were running out of time! Stephanie from a pedagogy perspective said it's also not about the speed of the answer so the timer only is there to add minor urgency.

bernhardccodev:
Hi @burrage, this is a legit bug from my last optimization. When I optimized the code, there was this condition `droppedStone === this.getCorrectTargetStone()` that ensures the feedback audio will play if the word spelling was completed. This condition was accidentally removed by me when I refactor some of the methods into one. So this PR is just re-adding of that condition so it would work as it is intended.

burrage:
Ah gotcha @bernhardccodev ! Thanks for that full context! That's my bad :-)

PR #1194, opened 2024-07-25
test: optimization on feedback text effect in gameplay screen

ashwinnair-chimple:
optimization on feedback text effect in gameplay screen disabled particle animation to test performance on low configuration devices


ashwinnair-chimple (commented):
optimization on feedback text effect in gameplay screen disabled text particle effect for testing performance in low config devices

PR #1206, opened 2024-08-15
optimize: [fm-225] - Optimize Background Implementation using Composition approach

bernhardccodev:
- Created a new way for initializing Background through the use of composition approach.
- Fixed levelNumber undefined in data this.levelData used for determining what background to display on every 10 levels of the game.


ashwinnair-chimple:
@bernhardccodev Lets confirm with Ashish and Akshat and then merge this PR regarding the levelNumber.

src/util/util.ts
ashwinnair-chimple, line 1: @bernhardccodev we already have utils.ts file in common folder lets keep this features inside that... also keep the common features in common folder and helpers in helper folder

bernhardccodev: Hi @ashwinnair-chimple , that utils.ts you're talking about is in a class util. I think moving my new utils.ts to that util class would defeat the purpose of what we are trying to achieve by minimizing unnecessary class creations.

bernhardccodev: @ashwinnair-chimple I see that there are other functions declared in the common/utils.ts. I'll move my utils in there.

bernhardccodev: Changes applied, consolidated all reusable methods from util/utils.ts and common/common.ts into common/utils.ts. Also moved the constant variables from common/utils.ts into constant/index.ts

ashwinnair-chimple (approved):

PR #1218, opened 2024-08-22
CR-99 : Content version Update

Nikhil77sutara:
-  Updated the service worker.js file for content version updates.
- Updated the feed the monster main file with content version updates.
- Removed the window.reload on cache complete events. Instead of this we are starting the ame by calling the start-scene functions.
- We have updated the service worker So, that we are caching the language content with language name in cache storage.
- We have also added exported DataModal as  a Class variable.


dz4va:
@Rajesh1041 @Nikhil77sutara I see that most of the audio files also have .gz added too! I everything correct there?

Nikhil77sutara:
> @Rajesh1041 @Nikhil77sutara I see that most of the audio files also have .gz added too! I everything correct there?

@dz4va I just updated my branch from develop and noticed these changes as well. Since these files are part of the build, I think everything is correct, and it shouldn't be a problem.

dz4va:
Awesome, thanks for the clarification, @Nikhil77sutara , I have approved the commit 🎉

dz4va (approved):

PR #1335, opened 2024-09-20
Feat: revamp feed the monster text design and animation (FM-272)

woodsjohn-codev:
# Changes
- Update the title text effect in the start screen
- Created a separate div for the title text
- Created a reusable function to apply font so that feedback and start scene title share the same code

# How to test
- Make sure start screen title is working properly or displaying properly in all languages.

Ref: [FM-272](https://curiouslearning.atlassian.net/browse/FM-272)

[FM-272]: https://curiouslearning.atlassian.net/browse/FM-272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/data/custom-fonts.ts
```diff
@@ -0,0 +1,7 @@
+export const feedbackCustomFonts: { [language: string]: string } = {
+  // english: "Consolas", // sample. if in future we need to customize feedback text in specific language
+};
+
+export const titleTextCustomFonts: { [language: string]: string } = {
```
bernhardccodev, line 5: These are good idea but why can we use this right now? Why can't we put the defaults we are using?

It's just doesn't make sense that you made a function to change the fonts but how are we using the original font? why can't we just call it here and have all the font changing be done here?

src/sceneHandler/index.ts
```diff
@@ -168,6 +171,7 @@ export class SceneHandler {
     this.showLoading();
     this.dispose(changeSceneRequestFrom);
     setTimeout(() => {
+      this.titleTextElement.style.zIndex = "-1";
```
bernhardccodev, line 174: What is the point of this.titleTextElement.style.zIndex = "-1";
Why is this only called before switching to gameplay and level selection scene?
Why not include this in level-end scene? If not due to gameplay, should we call it twice since you already called it in level selection?

src/constants/index.ts
```diff
@@ -23,6 +23,7 @@ export const GameScene1 = "GameScene1";
 export const EndScene1 = "EndScene1";
 export const FONT_BASE_PATH = './assets/fonts/';
 export const feedbackTextDefault = "Atma-SemiBold";
+export const titleTextDefaultFont = "Atma-SemiBold";
```
bernhardccodev, line 26: You keep creating functions that will handle the font change but you are not using those.

Please utilize the functions you have created instead of declaring them as a constant.

src/components/feedback-particle-effect/feedback-text-effects.ts
bernhardccodev, line 1: rename this file to index

src/common/utils.ts
```diff
@@ -169,13 +169,54 @@ export const toggleDebugMode = (toggleBtn: HTMLElement): void => {
...
+    style.id = styleId;
+    style.textContent = `
+      @font-face {
+        font-family: '${fontName}';
+        src: url('${fontPath}') format('truetype');
+        font-weight: normal;
+        font-style: normal;
+      }
+    `;
+    document.head.appendChild(style);
```
bernhardccodev, line 215: This function is being used by feedback-text-effects and start-scene.ts, so are we creating two elements?
Do we always append to the document.head every time feedback text loaded?

src/components/feedback-particle-effect/feedback-text-effects.ts
```diff
@@ -22,37 +22,26 @@ export class FeedbackTextEffects {
...
   public wrapText(text: string): void {
     if (!this.isFeedbackElementAvailable()) return;
-
     this.feedbackTextElement.textContent = text;
+
+    // Dynamically adjust the font size based on the length of the text
+    if (text.length >= 12) {
+      this.feedbackTextElement.style.fontSize = "30px";
+    } else {
+      this.feedbackTextElement.style.fontSize = ""; // Reset to default font size (or specify default if needed)
```
bernhardccodev, line 42: this can be simplified in one line, try ternary.

src/common/utils.ts
woodsjohn-codev: @bernhardccodev no. i created a checker for that if the font has been already added to the document head then this code will not execute https://github.com/curiouslearning/FeedTheMonsterJS/blob/82eb47cd558468fb7c00897842c8b5cae0a50ad2/src/common/utils.ts#L204-L216

bernhardccodev: why not include the check together with 
if (!fontName || !fontPath) {
    return;
  }

src/sceneHandler/index.ts
woodsjohn-codev: @bernhardccodev i remove this code to gameplay since it dont need it. I only need to hide the title when switching to level selection

src/components/feedback-particle-effect/feedback-text-effects.ts
woodsjohn-codev: thanks @bernhardccodev done.

src/components/feedback-particle-effect/feedback-text-effects.ts
woodsjohn-codev: done here. thanks @bernhardccodev

src/data/custom-fonts.ts
woodsjohn-codev: I used this file and set a font in every languages. pls check @bernhardccodev ebb9dad35ef40ffc1d67981b4b037c79a2f36384

src/constants/index.ts
ashwinnair-chimple: I agree with @bernhardccodev on this
@woodsjohn-codev please consider using the same.

src/data/custom-fonts.ts
bernhardccodev: You just made a commit to use this, right?

Also please use your codev and refrain from doing commits with your personal email

woodsjohn-codev: yes. thanks for the heads up. @bernhardccodev

bernhardccodev: @woodsjohn-codev how does titleTextCustomFonts and  feedbackCustomFonts works? Can we remove the default font declared in the constants `export const defaultFont = "Atma-SemiBold"`; if we have this two functions for fonts?

src/constants/index.ts
woodsjohn-codev: applied changes here. thanks guys

src/common/utils.ts
woodsjohn-codev: @bernhardccodev done here.

bernhardccodev: Let me rephrase my question: If this function needs to be triggered only once then why are me calling this in two separate files?

I just can't understand because we are checking if the font styles are already preloaded but feedback text will still call this method. Yes you have a condition that prevents it from running the condition again but that is why I am having a hard time due to it was already set but one file may still call this even though it has already set the font. Maybe we can optimize how this is being used so it will only be called once and no need for multiple files to trigger the font setting?

src/data/custom-fonts.ts
woodsjohn-codev: @bernhardccodev the purpose of titleTextCustomFonts and feedbackCustomFonts is to set a custom font incase the defaultFont dont support the other language if we add a new language in the future. so if ever theres a config inside of titleTextCustomFonts and feedbackCustomFonts it will use the font assigned. if not it will just use the defaultFont which is Atma

bernhardccodev: But can we remove the default fonts declared in constants and just use the default one through this function?

bernhardccodev: `private async loadFont() {
    const fontOptions = {
      customFonts: feedbackCustomFonts,
defaultFont: feedbackCustomFonts.default //instead of default
      lang: lang,
    };
    applyFontToElement(this.feedbackTextElement, fontOptions, FONT_BASE_PATH);
  } `

woodsjohn-codev: @bernhardccodev done i applied it already. thanks

src/common/utils.ts
woodsjohn-codev: @bernhardccodev i made an improvements here. i know load and apply the custom font in the feedthemonster.js file and removed the method i used on both files. this way it reduces the code.

bernhardccodev (approved):
Looks great to me.

ashwinnair-chimple (approved):

PR #1343, opened 2024-09-26
Feat: revamp feed the monster design (FM-272)

woodsjohn-codev:
# Changes
- fixed issue in the title screen when offline mode
- fix the FTM title in smaller screen device
- fix overlap in FTM title and dev button

# How to test
- Make sure that the title screen show properly in offline mode
- Make sure FTM title fits in one line 
- make sure deb button and title screen don't overlap to each other.

Ref: [FM-272](https://curiouslearning.atlassian.net/browse/FM-272)

[FM-272]: https://curiouslearning.atlassian.net/browse/FM-272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


feedTheMonster.ts
```diff
@@ -101,7 +100,11 @@ class App {
         element.style.fontFamily = fontFamily;
       }
     });
-   
+  }
+
+  private animateFTMText() {
+    this.titleTextElement.classList.remove("animate");
+    this.titleTextElement.classList.add("animate");
```
bernhardccodev, line 107: Why are we removing and then adding the same class?

feedTheMonster.ts
```diff
@@ -292,6 +296,7 @@ class App {
   }
 
   private reinitializeSceneHandler(dataModal: DataModal): void {
+    this.titleTextElement.style.zIndex = "2";
```
bernhardccodev, line 299: Why hard code here? Why not in css file?

src/sceneHandler/index.ts
```diff
@@ -196,6 +195,9 @@ export class SceneHandler {
     this.loadingScreen.initCloud();
     this.loading = true;
     document.getElementById("loading").style.zIndex = "3";
+    setTimeout(() => {
+      this.titleTextElement.style.zIndex =  "-1";
+    }, 800);
```
bernhardccodev, line 200: What is the reason adding another setTimeout here? Is this gonna be called once?

feedTheMonster.ts
woodsjohn-codev: this way we ensure that the title will re-animate when calling this function

src/sceneHandler/index.ts
woodsjohn-codev: So that it will have a seamless effect after the clouds finish animating. Without it, the FTM title will just hide instantly.

bernhardccodev: Then utilize CSS instead of creating another setTimeOut.

ashwinnair-chimple: Instead of zindex use property display:block/display:none, removing the element from DOM instead of just keeping it behind

feedTheMonster.ts
ashwinnair-chimple: no need of this line of code can be handled with pure css only, avoid using zindex with js until you have to use a DOM element on multiple scenarios. @woodsjohn-codev

feedTheMonster.ts
woodsjohn-codev: @bernhardccodev @ashwinnair-chimple pls check

src/sceneHandler/index.ts
woodsjohn-codev: @bernhardccodev @ashwinnair-chimple pls check

ashwinnair-chimple (approved):

bernhardccodev (dismissed):

feedTheMonster.ts
```diff
@@ -87,8 +87,9 @@ class App {
     }
     this.registerWorkbox();
 
-    this.loadTitleFeedbackCustomFont();
-    this.titleTextElement.classList.add("animate");
+    // load the title custom font before showing it
+    await this.loadTitleFeedbackCustomFont();
+    this.titleTextElement.classList.add("show");
```
ashwinnair-chimple, line 92: this is not required as by default class can be applied and remove when required

feedTheMonster.ts
```diff
@@ -292,6 +292,7 @@ class App {
   }
 
   private reinitializeSceneHandler(dataModal: DataModal): void {
+    this.titleTextElement.classList.remove(HIDE_ELEMENT_DELAY); // this is to remove hide delay class when scene resets.
```
ashwinnair-chimple, line 295: you can remove the class when reinitialization is called

ashwinnair-chimple (changes requested):
I have added few comments you can check that and then we can merge this PR

feedTheMonster.ts
bernhardccodev: +1

feedTheMonster.ts
```diff
@@ -1,7 +1,7 @@
 import * as Sentry from "@sentry/browser";
 import { getData, DataModal, customFonts } from "@data";
 import { SceneHandler } from "@sceneHandler";
-import { AUDIO_URL_PRELOAD, IsCached } from "@constants";
+import { AUDIO_URL_PRELOAD, HIDE_ELEMENT_DELAY, IsCached } from "@constants";
```
bernhardccodev, line 4: Remove this as well.

feedTheMonster.ts
bernhardccodev: +1

feedTheMonster.ts
woodsjohn-codev: please check @ashwinnair-chimple

bernhardccodev (approved):
All good on my end.

ashwinnair-chimple (approved):

PR #1346, opened 2024-09-27
Fix : Hindi Json revert

Nikhil77sutara:
# Changes
- Change 1

# How to test
- Its just for testing and we are reverting it.


Rajesh1041 (commented):
I am able to see changes in Tigragna changes there. Can you check?

Nikhil77sutara:
> I am able to see changes in Tigragna changes there. Can you check?

@Rajesh1041 I just commited hindi json changes. The above changes are coming becaues of comparing the branch i think.
It's fine I think we are already having those audios inside Tigragna.

PR #1353, opened 2024-10-01
Feat: subtle attention drawing animation on level icons (FM-271)

woodsjohn-codev:
# Changes
- Created a pulse effect on current level icon

# How to test
- Tap or click any part of the start screen to proceed to the next screen.
- On the level selection screen, ensure that the current level icon has a pulse effect animation.

Ref: [FM-271](https://curiouslearning.atlassian.net/browse/FM-271)

[FM-271]: https://curiouslearning.atlassian.net/browse/FM-271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/buttons/level-button.ts
+++ b/src/components/buttons/level-button.ts
@@ -28,7 +28,6 @@ export default class LevelBloonButton {
     private lockSize: number;
     private textFontSize: number;
     private pulse: boolean;
-    private isSpecial: boolean;
 
     constructor(
         canvas,
@@ -48,12 +47,10 @@ export default class LevelBloonButton {
         this.lockSize = canvas.height / 13;
         this.textFontSize = (this.size) / 6;
         this.pulse = false;
-        this.isSpecial = false;
     }
 
     setPulse(shouldPulse: boolean, isSpecial: boolean) {
         this.pulse = shouldPulse;
-        this.isSpecial = isSpecial;
     }
 
     isSpecialLevel(index){
--- a/src/components/buttons/level-button.ts
+++ b/src/components/buttons/level-button.ts
@@ -111,27 +111,36 @@ export default class LevelBloonButton {
     }
 
     private applyPulseEffect() {
-        const pulseDuration = 1500;
-        const maxOpacity = 0.5;
-        const baseColorRgba = '255, 255, 255';
-        const animationProgress = (Date.now() % pulseDuration) / pulseDuration;
-        const growPhase = animationProgress <= 0.7;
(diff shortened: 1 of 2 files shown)


src/components/buttons/level-button.ts
```diff
@@ -45,6 +46,11 @@ export default class LevelBloonButton {
         this.btnSize = this.bloonSize;
         this.lockSize = canvas.height / 13;
         this.textFontSize = (this.size) / 6;
+        this.pulse = false;
+    }
+
+    setPulse(shouldPulse: boolean, isSpecial: boolean) {
```
bernhardccodev, line 52: isSpecial is used here, do we still need to pass this?

src/scenes/level-selection-scene.ts
```diff
@@ -264,7 +264,12 @@ export class LevelSelectionScreen {
   };
 
   private drawLevel(levelBtn: any, gameLevelData: []) {
-    if (levelBtn.levelData.index + this.levelSelectionPageIndex <= this.data.levels.length) {
+    const currentLevelIndex = levelBtn.levelData.index + this.levelSelectionPageIndex;
+    const currentLevel = this.previousPlayedLevelNumber + 1;
+    
+    levelBtn.setPulse(currentLevelIndex === currentLevel, levelBtn.levelData.isSpecial);
```
bernhardccodev, line 270: Can we add this logic to levelBtn.draw?

 if (this.pulse) {   <------ replace this with the currentLevelIndex === currentLevel parameter, 
            this.applyPulseEffect();
        }
        
 I think this way it will me one less state to manage because you won't need to use a true or false flag.

src/components/buttons/level-button.ts
ashwinnair-chimple: where are we using setPulse method?

woodsjohn-codev: my bad. thanks for this. please check again @bernhardccodev

src/scenes/level-selection-scene.ts
woodsjohn-codev: thank you @bernhardccodev . I adjusted the code and remove the setpulse function since we dont need it anymore.

bernhardccodev (approved):
Reviewed and approved.

ashwinnair-chimple (approved):

PR #1355, opened 2024-10-02
Fix: Remove Next Level Icon after completing the Last Level of FTM (FM-281)

woodsjohn-codev:
# Changes
- Fix and hide the next level icon after completing the last level of FTM

# How to test
- Play the game until last level or toggle dev mode and select the last level and play
- Make sure next icon is not appearing in level end screen.
- Also test other languages just to make sure they work the same.

Ref: [FM-281](https://curiouslearning.atlassian.net/browse/FM-281)

[FM-281]: https://curiouslearning.atlassian.net/browse/FM-281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scenes/levelend-scene.ts
```diff
@@ -147,7 +148,10 @@ export class LevelEndScene {
       this.monster.update(deltaTime);
       this.closeButton.draw();
       this.retryButton.draw();
-      if (this.starCount >= 2) {
+      if (
+        this.starCount >= 2 &&
+        this.isLastLevel
```
bernhardccodev, line 153: do we need both condition for this condition or this.isLastLevel would suffice?

src/scenes/levelend-scene.ts
```diff
@@ -228,7 +232,11 @@ export class LevelEndScene {
       // pass same data as level is same
       this.switchToGameplayCB(gamePlayData, "LevelEnd");
     }
-    if (this.nextButton.onClick(x, y) && this.starCount >= 2) {
+    if (
+      this.nextButton.onClick(x, y) &&
+      this.starCount >= 2 &&
+      this.isLastLevel
```
bernhardccodev, line 238: this.nextButton.onClick(x, y) move this as the last condition so it won't fore the this.IsLastLevel is false.

woodsjohn-codev: Hi @bernhardccodev please check, again. thanks

src/scenes/levelend-scene.ts
woodsjohn-codev: We need both, because if we remove starCount, it means the user can go to the next level even if they only got 1 or 0 star. We need 2 stars in order to move to the next level.

bernhardccodev: I am sure about this. The task is to remove the next level on level-end scene if it is the last game level.

src/scenes/levelend-scene.ts
```diff
@@ -102,6 +102,7 @@ export class LevelEndScene {
     this.addEventListener();
     this.audioPlayer = new AudioPlayer();
     this.setupBg();
+    this.isLastLevel = this.data.levels.length === this.currentLevel + 1;
```
bernhardccodev, line 105: This is not the last level.

src/scenes/levelend-scene.ts
```diff
@@ -147,7 +148,7 @@ export class LevelEndScene {
       this.monster.update(deltaTime);
       this.closeButton.draw();
       this.retryButton.draw();
-      if (this.starCount >= 2) {
+      if (!this.isLastLevel && this.starCount >= 2) {
```
bernhardccodev, line 151: this is not correct, you only need to check if the game is the last level. The only reason is why you need this.starCount is because in line 105 is incorrect as you are not really getting or validating if it is the last level.
this.current is not an accurate indicator if the game level is last, it is only an indicator to the current level of the user has progressed to. I hope this is clear.

src/scenes/levelend-scene.ts
ashwinnair-chimple: there is nothing else to check other then the last level, may it be 0 star 1 star or 3 star there is an option to replay always but there should never be a next level button if it is the last level of the game. we can remove the starcount check from the condition.

src/scenes/levelend-scene.ts
ashwinnair-chimple: there is nothing else to check other then the last level, may it be 0 star 1 star or 3 star there is an option to replay always but there should never be a next level button if it is the last level of the game. we can remove the star count check from the condition.

ashwinnair-chimple: Apologies @woodsjohn-codev 
as discussed with Ashish and we do have a scenario where in any level if the stars are below 2 the next button wont be visible, resolving this comment.
Thanks

ashwinnair-chimple (approved):

ashwinnair-chimple:
@woodsjohn-codev can you include the build folder along with this PR

src/scenes/levelend-scene.ts
woodsjohn-codev: thank you @ashwinnair-chimple

ashwinnair-chimple (approved):

PR #1381, opened 2024-10-18
Fm 285

bernhardccodev:
# Changes
- Change 1

# How to test
- Test method 1

Ref: [<JIRA-CODE>](https://curiouslearning.atlassian.net/browse/<JIRA-CODE>)


src/game-state/game-data.ts
```diff
@@ -0,0 +1,74 @@
+import { DataModal } from "@data";
+import { SCENE_NAME_START } from "@constants";
+
+export class GameData {
```
miguelccodev, line 4: Can be shorten to
```
constructor(
    
    public data?: null | DataModal
    /* Canvas States */
    public canvas: null | HTMLCanvasElement
    public width: null | number
    public height: null | number
    public canavsElement = null;
    public context?: null | CanvasRenderingContext2D;
    ... // so on

) { }
```

if file length is a concern.

miguelccodev (commented):
Added initial reviews.

src/game-state/game-data.ts
bernhardccodev: For this part I am unsure. The constructor must remain empty because we are initializing this to a singletonInstance of gameStateService which doesn't have any data in the first place. The data or values will come once `gameStateService.setDefaultGameStateValues(data: DataModal,
        canvas: HTMLCanvasElement,
        canavsElement: HTMLCanvasElement)` has been called in the scene-handler.ts.

PR #1382, opened 2024-10-21
feat: FM-285 Game State Service Implementation

bernhardccodev:
# Changes
-Implemented Game State Service where the handling of game data and pub-sub events are handled and connected.
-Implemented Pub/Sub pattern for event-driven events.
-Created game data class for holding all the game values of the FTM.
-Added Data Access Object to properly extract the data from game data.
-LevelSelection and Level-End-Scene by refactored the callback to the gameplay-scene and replacing it with the gameStateService.
-SceneHandler -  In the constructor, this is where I called the method to update the game state data with the data passed to scene-handler from FeedTheMonster.ts to make it accessible to all.
-SceneHandler -  Refactor switchSceneToGameplay callback. Before this callback has another parameter which is the gamePlayData that is being passed from LevelSelection and LevelEnd-Selection. Now I removed it and point it to the gameStateService. The only remaining parameters in this method is the call backs to allow the gameplay-scene to switch to another scene.
-GamePlayScene - Uses GameStateServiceDao to load the initial data.
-GamePlayScene - Subscribes to pause event.
-GamePlayScene - Removed Trail Effect values for handling when to render trail-effects.
-GamePlayScene - Publishes using gameStateService when to trigger trailEffect.
-Trail-Particle - Moved the flag to trigger the trail effect from gameplay-scene to this class.
-Trail-Particle - Connected to gameStateService via subscribe to listen when should the trail effect to render.
-Stone-Handler - Has subscription to listen to pause-events
-Pause-Button Component - Publishes a trigger to set pause to TRUE.
-Pause-Popup - Publishes a trigger to set pause to FALSE when cancel button is clicked.
-Level-End-Scene - Patch fix to address the undefined issue of 'this' reference.

# How to test
- This is an overhaul to how the game passes data so play the app from start to end.

Ref: Ref: [<JIRA-CODE>](https://curiouslearning.atlassian.net/browse/FM-285)


miguelccodev (changes requested):
Add review comments

miguelccodev (approved):
Looks good! Approved.

woodsjohn-codev (approved):

ashwinnair-chimple (approved):

PR #1394, opened 2024-10-28
feat: FM-301 Refactor the End Scene UI from canvas to HTML, using background via BackgroundComponent & Rive Monster Component to load the Rive monster.

ashwinnair-chimple:
# Changes
- Refactored the Level End Scene from Canvas to Html
- Used Background Component and Rive Monster Component to render the elements on Level End Screen
- Refactored some assets and added that on gameplay screens
- Positioned elements on Gameplay and Level End Screen

# How to test
- Open the app and select any level play the gameplay screen and complete each levels and go back to levels or move ahead to next level.

Ref: [FM-301](https://curiouslearning.atlassian.net/browse/FM-301)

[FM-301]: https://curiouslearning.atlassian.net/browse/FM-301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scenes/levelend-scene.ts
```diff
@@ -84,131 +71,81 @@ export class LevelEndScene {
...
+    this.renderStarsHTML();
+  }
 
   switchToReactionAnimation = () => {
     if (this.starCount <= 1) {
       if (isDocumentVisible()) {
         this.audioPlayer.playAudio(AUDIO_LEVEL_LOSE);
       }
-      this.monster.changeToSpitAnimation();
+      // this.monster.changeToSpitAnimation();
```
bernhardccodev, line 107: I think we can completely removed this since we are no longer using Canvas.

src/scenes/levelend-scene.ts
```diff
@@ -84,131 +71,81 @@ export class LevelEndScene {
...
     } else {
       if (isDocumentVisible()) {
         this.audioPlayer.playAudio(AUDIO_LEVEL_WIN);
         this.audioPlayer.playAudio(AUDIO_INTRO);
       }
-      this.monster.changeToEatAnimation();
+      // this.monster.changeToEatAnimation();
```
bernhardccodev, line 113: Same here, this can be completely removed.

src/scenes/levelend-scene.ts
```diff
@@ -84,131 +71,81 @@ export class LevelEndScene {
...
-    this.background = await createBackground(
-      this.context,
-      this.width,
-      this.height,
-      BG_GROUP_IMGS,
-      draw
-    );
-  };
+  showLevelEndScreen() {
+
```
bernhardccodev, line 91: Remove these two spaces line 91 and 92.

src/scenes/levelend-scene.ts
```diff
@@ -84,131 +71,81 @@ export class LevelEndScene {
...
-      this.closeButton.draw();
-      this.retryButton.draw();
-      if (this.isLastLevel) {
-        this.nextButton.draw();
-      }
+    // Draw monster and buttons
+    this.closeButton.draw();
+    this.retryButton.draw();
+    if (this.isLastLevel) {
+      this.nextButton.draw();
```
bernhardccodev, line 124: Optional: This is only a suggestion but I think it would be much cleaner if we could write it like `this.isLastLevel && this.nextButton.draw()`, instead of using IF condition.

public/index.css
```diff
@@ -458,7 +457,43 @@ img {
...
   background-image: url("./assets/images/bg_v01.webp");
 }
-
+#levelEnd{
+  display: none;
+  height: 100vh;
+  width: 100%;
+  position: fixed;
+  z-index: 7;
+  background-image: url(./assets/images/WIN_screen_bg.webp);
```
bernhardccodev, line 466: Question, our level end scene has also seasonal backgrounds. Do we also handle that seasonal changes in here or somewhere?

bernhardccodev (commented):
Please see comments, thank you.

public/index.css
ashwinnair-chimple: As the level end background is a static background so we can keep it separately in css, but all the seasonal background is already handled from background component and that will be visible through the gameplay background which is already rendered so no need to handle it separately.

src/scenes/levelend-scene.ts
ashwinnair-chimple: this is commented to handle the interactive animation on the levelend screen, I will add comments for more clarity.

bernhardccodev (approved):
All good on my end and thank you for answering my questions.

woodsjohn-codev (approved):

miguelccodev (approved):
Looks good to me.

Approved.

PR #1397, opened 2024-10-30
feat: FM-302 Create the Level End Scene UI in HTML, supporting DOM ma…

ashwinnair-chimple:
Create the Level End Scene UI in HTML, supporting DOM manipulation and other functionality.

# Changes
- The Level End Scene UI is built using HTML.
- DOM manipulation is supported for dynamic content display.
- The scene works consistently across all devices.

# How to test
- Open the app and select any level play the gameplay screen and complete each levels and go back to levels or move ahead to next level or play the level again.

Ref: [FM-302](https://curiouslearning.atlassian.net/browse/FM-302)

[FM-302]: https://curiouslearning.atlassian.net/browse/FM-302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/scenes/levelend-scene.ts
+++ b/src/scenes/levelend-scene.ts
@@ -70,7 +70,7 @@ export class LevelEndScene {
     this.starCount = starCount;
     this.currentLevel = currentLevel;
     // Subscribe to the LEVEL_END_BACKGROUND_TOGGLE event
-    gameStateService.subscribe(gameStateService.EVENTS.LEVEL_END_BACKGROUND_TOGGLE, this.toggleLevelEndBackground);
+    this.toggleLevelEndBackground(true);
     this.isLastLevel =
       this.currentLevel !==
       this.data.levels[this.data.levels.length - 1].levelMeta.levelNumber &&
@@ -93,7 +93,7 @@ export class LevelEndScene {
 
   // Call this method where necessary to show the background when level ends
   showLevelEndScreen() {
-    gameStateService.toggleLevelEndBackground(true); // Publish show event
+    this.toggleLevelEndBackground(true); // Publish show event
     // Render the stars dynamically based on the star count
     this.renderStarsHTML();
   }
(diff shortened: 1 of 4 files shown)


src/scenes/levelend-scene.ts
```diff
@@ -70,27 +69,31 @@ export class LevelEndScene {
...
     this.renderStarsHTML();
   }
-
-  showLevelEndScreen() {
-    // Make the levelEnd element visible by setting display to block
+  // Method to show/hide the Level End background
+  toggleLevelEndBackground = (shouldShow: boolean) => {
     if (this.levelEndElement) {
-      this.levelEndElement.style.display = 'block';
+      this.levelEndElement.style.display = shouldShow ? 'block' : 'none';
```
bernhardccodev, line 90: Now that I can see the whole logic, I think it can be simplified.
My suggestion would be to;

Your method toggleLevelEndBackground will be retained.
```
toggleLevelEndBackground = (shouldShow: boolean) => {
    if (this.levelEndElement) {
      this.levelEndElement.style.display = shouldShow ? 'block' : 'none';
    }
  };
```
In the constructor since we need this to be set to block everytime levelend-scene loads.
```
constructor() {
this.toggleLevelEndBackground(true)
}
```
 In the level-end-scene I have a method handlePublishEvent that I used to reuse for publishing events to game state this is where you can hook up the removal.
```
private handlePublishEvent(shouldShowLoading: boolean, gamePlayData = null) {
    gamePlayData && gameStateService.publish(gameStateService.EVENTS.GAMEPLAY_DATA_EVENT, gamePlayData);
    gameStateService.publish(gameStateService.EVENTS.SCENE_LOADING_EVENT, shouldShowLoading);
    this.toggleLevelEndBackground(!shouldShowLoading) ---> your toggleLevelEndBackground for removal

  }
```
If you would check handlePublishEvent method will always be expecting a true for shouldShowLoading because this is always called whenever we are expecting a changed scene from levelend-scene to either gameplay or level selection so this is the best place to use or handle the css hiding.
With this you can pretty much disregard the game-state and scene-handler changes.

bernhardccodev (commented):
I have some comments, please consider this suggestion.

src/scenes/levelend-scene.ts
ashwinnair-chimple: oh wait, The reason to use the changes in scene-handler is cause there is a delay when hiding the background from level end screen, So I was making sure that it only hides when the screen redirects and not before that. let me share a recording with you @bernhardccodev

ashwinnair-chimple: https://github.com/user-attachments/assets/793ad0ef-1f62-4928-8ff9-0e19dba1edf9

@bernhardccodev for you reference. by adding the check in scenehandler i only hide the background when the screen is redirecting and not before that.

bernhardccodev (approved):
Looks good to me! Approved.

woodsjohn-codev (approved):

miguelccodev (approved):
Approved

PR #1419, opened 2024-11-11
Feat: Refactor the Level Indicator from Canvas to HTML (FM-311)

bernhardccodev:
# Changes
- Created Base HTML for reusable class for creating, appending div templates and removing them.
-Created Level Field class where it handles all relating to rendering and extended using Base HTML.
-Code Clean up for Level Indicator by removing all canvas related drawings of the previous level indicator.
-Created Unit testing for both Base HTML and Level Field Component.

# How to test
- Start The FTM
-Select any game level.
-During game play, notice the level indicator on top left, it should update whenever you finish a stage in a game.

Ref: [FM-311](https://curiouslearning.atlassian.net/browse/FM-311)

[FM-311]: https://curiouslearning.atlassian.net/browse/FM-311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/level-indicator.ts
```diff
@@ -1,78 +1,29 @@
...
-                    (this.canvas.width * 0.35) / 10,
-                    (this.canvas.height * 0.09) / 3
-                );
-            }
-        }
-    }
 
     public dispose() {
         this.unregisterEventListener();
+        this.levelBarIndicator.removeLevelField();
```
miguelccodev, line 26: can just call `destroy` directly.

miguelccodev (changes requested):
just one minor comment.

miguelccodev (approved):

woodsjohn-codev (approved):

src/components/baseHTML/BaseHTML.ts
ashwinnair-chimple, line 1: suggestion to use consistent naming convention through out so you can use camelCase baseHtml/baseHtml.ts

src/components/baseHTML/BaseHTML.ts
```diff
@@ -0,0 +1,55 @@
+export interface BaseHtmlOptions {
+  selectors: {
+    root: string;
+    [key: string]: string;
+  }
+}
+
+export interface HTML_TEMPLATE {
```
ashwinnair-chimple, line 8: using HtmlTemplate will be better than HTML_TEMPLATE

src/components/baseHTML/BaseHTML.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/*
+* BaseHTML class used as a base for creating classes that needs to create and append new HTML elements.
+* This class also serves a an ease to easily manipulate the created HTML element by passing a template and
+* adding a method for ease of removing it.
+*/
+export class BaseHTML {
+  private isRendered: boolean = false;
+  private id: string;
+  private template: HTML_TEMPLATE;
```
ashwinnair-chimple, line 20: using HtmlTemplate will be better than HTML_TEMPLATE

src/components/baseHTML/BaseHTML.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.render();
+    //Add more logic here if needed.
+  }
+
+  private render() {
+    if (this.isRendered) return;
+    const { root } = this.options.selectors;
+    const rootEl = document.querySelector(root);
+    const templateString = this.template(this.id);
+    rootEl.insertAdjacentHTML('beforeend', templateString);
```
ashwinnair-chimple, line 42: add check for rootE1 & templateString what if it is null/undefined before inserting

src/components/baseHTML/BaseHTML.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.isRendered = true;
+  }
+
+  public getElements(className) {
+    return document.querySelectorAll(className) as NodeListOf<HTMLImageElement>
+  }
+
+  public destroy() {
+    document.getElementById(this.id).remove();
```
ashwinnair-chimple, line 51: add check if element exist before directly removing

ashwinnair-chimple (commented):
add few comments

src/components/level-field/level-field-component.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+
+export const LEVEL_FIELD_LAYOUT = (id: string) => {
+  return (`
+    <div id="${id}" class="level_content-wrapper">
+      <div class="level-field">
+        <img class="level-background" src="${LEVEL_INDICATOR}"/>
+        <div class="bar-level-wrapper">
+          <img class="bar-level" src="${BAR_EMPTY}" />
```
ashwinnair-chimple, line 24: I have modified this to more optimized way also add ALT property

<div class="bar-level-wrapper">
  ${Array(5).fill(`<img class="bar-level" src="${BAR_EMPTY}" alt="any content/>`).join('')}
  </div>

src/components/baseHTML/BaseHTML.ts
bernhardccodev: renaming it instead to base-html as Git cannot tell the difference of BaseHtml to baseHtml.

src/components/baseHTML/BaseHTML.ts
bernhardccodev: Understandable but in this case an ID is a hard requirement in using baseHtml so it is guaranteed to have an ID otherwise it will error out during development.

ashwinnair-chimple (approved):
Approved

src/components/baseHTML/BaseHTML.ts
bernhardccodev: However I did changed it to `this.isRendered && document.getElementById(this.id).remove();` this way using the destroy method more than once would avoid any error.

PR #1454, opened 2024-12-06
feat: Converted Start Scene background from canvas to HTML. (fm-288)

bernhardccodev:
# Changes
- Replaced Start Scene background from Canvas into HTML background.

# How to test
- Start the application
- In the very first screen before you press enter, check if the background loads.

Ref: [FM-288](https://curiouslearning.atlassian.net/browse/FM-288)

[FM-288]: https://curiouslearning.atlassian.net/browse/FM-288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/scenes/start-scene.ts
+++ b/src/scenes/start-scene.ts
@@ -1,18 +1,23 @@
 import { Monster, AudioPlayer, BackgroundHtmlGenerator } from "@components";
-import { PlayButton } from "@buttons";
+import { PlayButtonHtml } from '@components/buttons';
+import { BaseButtonComponent } from '@components/buttons/base-button-component/base-button-component';
+import { RiveMonsterComponent } from "@components/riveMonster/rive-monster-component";
 import { DataModal } from "@data";
 import {
   StoneConfig,
   toggleDebugMode,
   Utils,
+  pseudoId,
+  lang
 } from "@common";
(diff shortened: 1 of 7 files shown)


src/scenes/start-scene.ts
```diff
@@ -81,13 +78,15 @@ export class StartScene {
...
-    );
+    // Determine the background type based on the level number using the static method.
+    //Level 1 will be used as a default background for Start Scene.
+    const selectedBackgroundType = BackgroundHtmlGenerator.createBackgroundComponent(1);
+
+    // Apply the logic to update the HTML or visual representation of the background
+    const backgroundGenerator = new BackgroundHtmlGenerator();
+
+    // Dynamically update the background based on the selected type
+    backgroundGenerator.generateBackground(selectedBackgroundType);
```
woodsjohn-codev, line 89: if possible can we add one test case for start scene? thanks!

src/scenes/start-scene.ts
ashwinnair-chimple, line 1: make sure the canvas is removed and play button is handled through HTML, #canvas should only be visible on gameplay screen. and only rivecanvas should be visible on start screen.
On click of play button FM-291 should be used to handle click event and other functional changes.

PR #1456, opened 2024-12-09
feat: Updated Start Scene from canvas implementation to HTML.

bernhardccodev:
# Changes
- Replaced the start scene background from Canvas to HTML implementation.
- Updated Timer component creation, originally using a background class, now points to the ID background. So it won't bug out due to the implementation of Start Scene HTML background.
- Updated base button component by adding a _destroy method.
- Added test cases for background component and updated test case of base button by adding the destroy test scenario.

# How to test
- Run FTM app on browser.
- In start scene, press F12 to check if the rendered Play Button and Background are HTML implementation by viewing them on the DOM tree.
- Press Play and select any game level.
- In the game level, there should be a timer rendered.

Ref: [FM-288](https://curiouslearning.atlassian.net/browse/FM-288)

[FM-288]: https://curiouslearning.atlassian.net/browse/FM-288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/buttons/base-button-component/base-button-component.ts
+++ b/src/components/buttons/base-button-component/base-button-component.ts
@@ -92,7 +92,7 @@ export class BaseButtonComponent {
     return this.element;
   }
 
-  public _destroy() {
+  public destroy() {
     const element = document.getElementById(this.id);
     if (element) element.remove();
   }
--- a/src/components/buttons/base-button-component/base-button-component.ts
+++ b/src/components/buttons/base-button-component/base-button-component.ts
@@ -93,7 +93,6 @@ export class BaseButtonComponent {
   }
 
   public _destroy() {
-    console.log('this.id ', this.id)
     const element = document.getElementById(this.id);
     if (element) element.remove();
   }
--- a/src/components/buttons/base-button-component/base-button-component.ts
+++ b/src/components/buttons/base-button-component/base-button-component.ts
@@ -10,14 +10,15 @@ export interface ButtonOptions {
 }
 
 export class BaseButtonComponent {
+  protected id: string;
   protected element: HTMLElement;
   protected audioPlayer: AudioPlayer;
   private clickListener: (event: Event) => void;
 
   constructor(options: ButtonOptions) {
+    this.id = options.id;
     this.audioPlayer = new AudioPlayer();
     const existingElement = document.getElementById(options.id);
-
     if (existingElement) {
       this.element = existingElement;
     } else {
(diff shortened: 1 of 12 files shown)


woodsjohn-codev (approved):

src/components/buttons/base-button-component/base-button-component.ts
```diff
@@ -90,4 +91,9 @@ export class BaseButtonComponent {
   public getElement(): HTMLElement {
     return this.element;
   }
+
+  public _destroy() {
```
miguelccodev, line 95: rename to `destroy()`.

Let's keep this consistent, eventually all html components will extend the base html one.

src/scenes/start-scene.ts
```diff
@@ -158,4 +163,17 @@ export class StartScene {
     this.pwa_install_status = event;
     localStorage.setItem(PWAInstallStatus, "false");
   };
+
+  /*Note: This was from PlayButton canvas class onclick method.*/
+  private logTappedStartFirebaseEvent() {
```
miguelccodev, line 168: Add a unit test that expects `sendTappedStartEvent` to have been called.

miguelccodev (changes requested):
Added comments.

src/components/buttons/play-button/play-button.scss
```diff
@@ -0,0 +1,44 @@
...
+@media (min-width: 768px) {
+  #play-button {
+    max-width: 200px;
+    max-height: 200px;
+  }
+}
+
+//lg
+@media (min-width: 992px) {
+
```
ashwinnair-chimple, line 33: remove all the unused media queries to keep the code clean

src/components/buttons/play-button/play-button.scss
```diff
@@ -0,0 +1,44 @@
+.game-scene {
+  #play-button {
+      position: absolute;
+      top: 23vh;
+      left: 50%;
+      right: 50%;
```
ashwinnair-chimple, line 6: Only left: 50% is needed when using transform: translate(-50%, -50%)
right: 50% can be remobed

ashwinnair-chimple (commented):
added comments

miguelccodev (approved):

ashwinnair-chimple (approved):
Approved

PR #1481, opened 2025-01-03
fix: FM-353 fix pulse issue in level select scene (FM-353)

woodsjohn-codev:
# Changes
- Fix the issue when the pulse is animating even there's no balloon or level

# How to test
- Check and make sure pulse is animating in the last level if the last level is unlocked.
- Check this fix in other language that has a few levels like georgian

Ref: [FM-353](https://curiouslearning.atlassian.net/browse/FM-353)

[FM-353]: https://curiouslearning.atlassian.net/browse/FM-353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/buttons/level-button.ts
+++ b/src/components/buttons/level-button.ts
@@ -64,7 +64,6 @@ export default class LevelBloonButton {
         levelSelectionPageIndex,
         unlockLevelIndex,
         gameLevelData,
-        totalGameLevels
     ) {
         this.context.drawImage(
             this.levelData?.balloonImg,
@@ -96,7 +95,6 @@ export default class LevelBloonButton {
             levelSelectionPageIndex,
             unlockLevelIndex,
             gameLevelData,
-            totalGameLevels
         );
     }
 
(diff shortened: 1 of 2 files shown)


src/scenes/level-selection-scene.ts
```diff
@@ -258,6 +258,12 @@ export class LevelSelectionScreen {
     }
 
     for(let btn of this.levelButtons) {
+      // Check if the level exists before allowing click
+      const levelIndex = btn.levelData.index + this.levelSelectionPageIndex;
+      if (levelIndex > this.data.levels.length) {
+        continue; // Skip this button if level doesn't exist
+      }
```
bernhardccodev, line 265: Hi @woodsjohn-codev , why not prevent the excess creation instead of adding a flag to prevent the buttons that should not be there in the first place?

src/components/buttons/level-button.ts
```diff
@@ -209,6 +209,12 @@ export default class LevelBloonButton {
         unlockLevelIndex: number,
         callBack
     ) {
+        // Check if this level exists
+        const levelIndex = this.levelData.index + levelSelectionPageIndex + 1;
+        if (levelIndex > this.totalGameLevels) {
```
ashwinnair-chimple, line 214: @woodsjohn-codev what is the purpose of the click when we dont have buttons/levelballons can we just control the loop that generates the buttons?

bernhardccodev (approved):

ashwinnair-chimple (approved):
Approved

PR #1487, opened 2025-01-07
fix: Fix for Cloud animation plays for every click on the screen (FM-352).

bernhardccodev:
# Changes
- Moved the onclick event that allows the user to proceed to level selection via onclick event to a new div container using base html.
- Added new feature to base html that sanitize if the element already exists. This is can be used by passing true to onSanitize
- Updated test case for base HTML.

# How to test
- Run FTM locally
- On start scene you still should be able to proceed to level selection by clicking anywhere.
- Also regress testing by opening the app and then going back to FTM container, then open the FTM again
- You should be able to use the application with triggering the clouds loading bug.

Ref: [FM-352](https://curiouslearning.atlassian.net/browse/FM-352)

[FM-352]: https://curiouslearning.atlassian.net/browse/FM-352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bernhardccodev:
I will close this and re-apply my fixes by branching out of the develop branch.

PR #1534, opened 2025-02-06
fix: build race condition

miguelccodev:
# Changes
- circle ci config, to require build to finish before deploying to s3
- added coverage to git ignore

# How to test
- observed in circle ci dashboard, build should be finished first before deploying.


ashwinnair-chimple (approved):
approved

PR #1560, opened 2025-02-21
feat: integrated feaeture flags (EN-33)

miguelccodev:
# Changes
- feat: added @curiouslearning/features dependency
- feat: integrated en-33 feature

# How to test
- When no id is present, clicking start should skip level selection and go straight to gameplay (quick start)
- When id is present, clicking start should go through default screens
- When offline, it should retain loaded feature
- When offline without stored feature flags, it should default to false (default behavior)

Ref: [EN-33](https://curiouslearning.atlassian.net/browse/EN-33)

Note: unit tests done separately in features module

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@capacitor/android": "^4.4.0",
         "@capacitor/core": "^4.4.0",
+        "@curiouslearning/features": "^1.2.5",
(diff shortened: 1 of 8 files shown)


bernhardccodev (approved):
Hi @miguelccodev, everything looks good to me! For the To Do clean up of gamePlayData, can we create a ticket for that one?

woodsjohn-codev (approved):

PR #1565, opened 2025-02-28
feat: Added background evolution for different monster phases (Fm-375).

bernhardccodev:
# Changes
- Created Phase Background.ts for handling the evolution background.
- Disabled Seasons Background in Gameplay scene.
- Temporary replaced Season background with Phase Background.

# How to test
- Build FTM
- Start the FTM game
- Play from levels 1 - 3 and aim for 3 scores.
- By playing the game with perfect score, after level 3 the monster should evolved.
- Next level, the background should be different and changed as the monster evolves.

Ref: [FM-375](https://curiouslearning.atlassian.net/browse/FM-375)

[FM-375]: https://curiouslearning.atlassian.net/browse/FM-375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/constants/index.ts
```diff
@@ -157,3 +157,8 @@ export const SCENE_NAME_LEVEL_END = "LevelEnd";
 
 //Levels
 export const SPECIAL_LEVELS = [5, 13, 20, 30, 42];
+
+//Evolving Phases Backgrounds
+export const PHASE_1_BG = './assets/images/placeholder_background_1.webp';
+export const PHASE_2_BG = './assets/images/placeholder_background_2.webp';
+export const PHASE_3_BG = './assets/images/placeholder_background_3.webp';
```
dz4va, line 164: I thought of having an enum here to make things more consistent and contained? @bernhardccodev 
Something like:

```
export enum EvolutionBackgroundPhaseAssets {
  Phase1 = './assets/images/placeholder_background_1.webp',
  Phase2 = './assets/images/placeholder_background_2.webp',
  Phase3 = './assets/images/placeholder_background_3.webp',
}
```

And in the phases-background.ts we would use it like:

`const bgImage: string = EvolutionBackgroundPhaseAssets.Phase1;`

This way we wouldn't even need to have a comment above these 3 PHASE_BG const variables, and we would get a dropdown after typing `EvolutionBackgroundPhaseAssets.` as a plus which would only show however many phases are available in total

+ an enum can be used just like a const Typescript object without initialization `EvolutionBackgroundPhaseAssets.Phase1`

Please let me know your thoughts on this, @bernhardccodev

src/components/background/phases-background.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    section.append(backgroundImg);
+    return section;
+  }
+
+  // Clears only the content within the #background-elements container
+  public clearBackgroundContent(): void {
+    const backgroundElementsContainer = document.getElementById(
+      "background-elements"
+    );
```
dz4va, line 71: Duplicated code, let's keep it DRY as much as possible, we are already grabbing the value in:

```
public generateBackground(phase: number): void {
    const backgroundElementsContainer = document.getElementById(
      "background-elements"
    );

    if (!backgroundElementsContainer) return;

    const assets = this.getAssetsForSeason(phase);
    this.clearBackgroundContent();

    const section = this.createBackgroundSection(assets);
    backgroundElementsContainer.appendChild(section);
  }
```

Why don't we set the "background-elements" html element as a member variable for this class and call it `BackgroundElementsContainer` to share it in both methods and possibly in other in the future

Thoughts on this @bernhardccodev ? 🎉

src/components/background/phases-background.ts
```diff
@@ -0,0 +1,76 @@
...
+      case 1:
+        return {
+          placeholder: PHASE_2_BG,
+        };
+      case 0:
+      default:
+        return {
+          placeholder: PHASE_1_BG ,
+        };
+    }
```
dz4va, line 45: With the Enum approach we could have something like this here:

`placeholder: Object.values(EvolutionBackgroundPhaseAssets)[phase] || EvolutionBackgroundPhaseAssets.Phase1,`

Since the phase numeric value would correspond to the Phase1, Phase2, and Phase3 as suggested down in the comments

Thoughts, @bernhardccodev ? 🙌🏽

dz4va (changes requested):

woodsjohn-codev (approved):
I have a minor comments

src/components/background/phases-background.ts
```diff
@@ -0,0 +1,76 @@
...
+  placeholder: string;
+}
+
+export class PhasesBackground extends BaseBackgroundComponent {
+  constructor() {
+    super("background");
+  }
+
+  // Generates and appends the background section for a given phase number.
+  public generateBackground(phase: number): void {
```
woodsjohn-codev, line 14: is it okayt if we add type safety here?
something like:

```
type PhaseNumber = 0 | 1 | 2;
public generateBackground(phase: PhaseNumber): void
```

src/components/background/phases-background.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const assets = this.getAssetsForPhase(phase);
+    this.clearBackgroundContent();
+
+    const section = this.createBackgroundSection(assets);
+    backgroundElementsContainer.appendChild(section);
+  }
+
+  // Retrieves assets based on the phase of monster.
+  private getAssetsForPhase(phase: number): BackgroundAssets {
```
woodsjohn-codev, line 29: if you set PhaseNumber above you can update this one:

`private getAssetsForPhase(phase: PhaseNumber)`

src/components/background/phases-background.ts
bernhardccodev: Hi @dz4va, for context this code is derived from the background.ts component to maintain consistency and uniformity across the project. But this makes sense, I will apply this.

src/components/background/phases-background.ts
bernhardccodev: You're saying to ensure the values will be within the intended numbers, right? Got it.

src/constants/index.ts
bernhardccodev: @dz4va Code changes applied, thank you!

src/components/background/phases-background.ts
bernhardccodev: @dz4va Code changes applied, thank you!

src/components/background/phases-background.ts
bernhardccodev: @woodsjohn-codev code changes applied, thank you!

src/components/background/phases-background.ts
bernhardccodev: @woodsjohn-codev code changes applied, thank you!

woodsjohn-codev (approved):

dz4va (approved):

PR #1573, opened 2025-03-11
feat: Rive of Phase 2 magnet monster (Fm 392)

bernhardccodev:
# Changes
- Updated the Rive assets of Phase 2, from template blue egg to ftm_monster_1_2_v03.riv.
- Fixed the issue in Rive implementation where IDLE animation overlaps with state machines in Gameplay-scene.
- Added a dynamic layout to handle the rive positioning during evolution.

# How to test
- Build and Play FTM.
- When reaching to total of 8 stars, the monster would evolved.
- The placement of the evolution animation should be on top of the background trees stump.
- Next level, it should render the phase two rive monster.
- It should be the new asset, a hatched egg monster should be rendered instead of a blue egg template.

Ref: [FM-392](https://curiouslearning.atlassian.net/browse/FM-392)

[FM-392]: https://curiouslearning.atlassian.net/browse/FM-392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ashwinnair-chimple (approved):
Approved!

PR #1592, opened 2025-03-18
feat: Reduced delay of reloading the puzzle and switching from gameplay to levelend (FM-434)

bernhardccodev:
# Changes
- Reduced the 4500 milliseconds to 3000 milliseconds before each puzzles resets and switches from gameplay to levelend whenever an incorrect answer was given.
- Added a logic that handles which delays to use 4500 or 3000 milliseconds depending on the correct answer given by the user.
- Fixed the IDLE overriding with the HAPPY animation at levelend-scene.

# How to test
- Start the FTM app
- Select any game level
- Play the game and give the correct or incorrect answer.
- There should be no long wait before each puzzles resets or switches to levelend scene.

Ref: [FM-434](https://curiouslearning.atlassian.net/browse/FM-434)


ashwinnair-chimple (approved):
approved

woodsjohn-codev (approved):
Approved

PR #1614, opened 2025-03-26
feat: pause the timer clock animation when the game is paused (Fm-448)

woodsjohn-codev:
# Changes
- Pause the timer and clock animation when the the game play is in pause

# How to test
- Select any level and play
- Press pause button
- Make sure the timer clock icon pauses as well and resume when hit resume or x button from the modal

Ref: [Fm-448](https://curiouslearning.atlassian.net/browse/Fm-448)


src/components/timer-ticking.ts
```diff
@@ -115,6 +116,28 @@ export default class TimerTicking extends EventManager {
         }, 0);
     }
 
+    /**
+     * Stops the clock rotation when the game is paused
+     */
```
ashwinnair-chimple, line 121: @woodsjohn-codev i have added a method on timer-ticking.ts `this.applyRotation(false); ` just pass false when you want to stop rotation and pass true when you want to rotate the clock

src/components/timer-ticking.ts
```diff
@@ -115,6 +116,28 @@ export default class TimerTicking extends EventManager {
...
+    public pauseRotation(): void {
```
ashwinnair-chimple, line 122: no need of this and the below methods

src/components/timer-ticking.ts
woodsjohn-codev: applied  update here @ashwinnair-chimple

src/components/timer-ticking.ts
woodsjohn-codev: applied  update here @ashwinnair-chimple

src/components/timer-ticking.ts
```diff
@@ -106,15 +110,61 @@ export default class TimerTicking extends EventManager {
...
+        const timerContainer = document.getElementById("timer-ticking");
+        if (timerContainer) {
+            timerContainer.classList.add('paused-animations');
+        }
+    }
+
+    /**
+     * Resumes the clock rotation if the timer is still running
+     */
+    public resumeRotation(): void {
```
bernhardccodev, line 135: Hi, I think `resumeRotation` and `pauseRotation` method are almost the same. I think we can follow the DRY principle here and maybe only have one method for controlling the rotation. Thoughts @woodsjohn-codev ?

bernhardccodev (changes requested):

src/components/timerHtml/timerHtml.scss
```diff
@@ -4,6 +4,10 @@
   width: 100%;
   display: flex;
 
+  /* Pause all animations when this class is added */
```
ashwinnair-chimple, line 7: can we remove this if not used

ashwinnair-chimple (approved):
approved

PR #1628, opened 2025-04-02
fix: Reverted back the original implementation for minY (FM-441)

bernhardccodev:
# Changes
- Reverted back to the original implementation of minY as FM-441 will now be a research ticket.

# How to test
- Run the app and play any game level.
- The rive monster should render at the stump or within.

Ref: [FM-441](https://curiouslearning.atlassian.net/browse/FM-441)


ashwinnair-chimple (approved):
approved

PR #1640, opened 2025-04-15
feat: Move Feedback Audio players from stone-handler to game puzzles class and create a reusable component for Feedback audio. (FM-468)

woodsjohn-codev:
# Changes
- Refactored puzzle logic into a modular, extensible architecture with separate components for different puzzle types and audio feedback, improving maintainability and testability.

# How to test
- Select any level and play
- make sure audio is still working as before

Ref: [FM-468](https://curiouslearning.atlassian.net/browse/FM-468)

[FM-468]: https://curiouslearning.atlassian.net/browse/FM-468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/scenes/gameplay-scene/gameplay-scene.ts
+++ b/src/scenes/gameplay-scene/gameplay-scene.ts
@@ -40,7 +40,12 @@ import {
   PreviousPlayedLevel,
   MONSTER_PHASES
 } from "@constants";
-import { WordPuzzleLogic } from '@gamepuzzles';
+import { 
+  BasePuzzleLogic, 
+  PuzzleFactory, 
+  FeedbackAudioHandler, 
+  FeedbackType 
+} from '@gamepuzzles';
 import gameStateService from '@gameStateService';
 import gameSettingsService from '@gameSettingsService';
 import { PAUSE_POPUP_EVENT_DATA, PausePopupComponent } from '@components/popups/pause-popup/pause-popup-component';
@@ -88,7 +93,8 @@ export class GameplayScene {
   puzzleTime: number;
   isDisposing: boolean;
   trailParticles: any;
-  wordPuzzleLogic: any;
+  puzzleLogic: BasePuzzleLogic;
+  feedbackAudioHandler: FeedbackAudioHandler;
   public riveMonsterElement: HTMLCanvasElement;
   public gameControl: HTMLCanvasElement;
   private unsubscribeEvent: () => void;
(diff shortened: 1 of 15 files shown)


src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -404,17 +411,17 @@ export class GameplayScene {
         trailX = newStoneCoordinates.x;
         trailY = newStoneCoordinates.y;
 
-        if (this.wordPuzzleLogic.checkIsWordPuzzle()) {
+        if (this.puzzleLogic.checkIsWordPuzzle()) {
```
bernhardccodev, line 414: Hi @woodsjohn-codev, for this one from 375 to 414, it should not be a simple replaced of name.
If you would review the whole scope of this is that this is one long if condition that determines if the logic should use word puzzle or letter.

This should be optimized as well because if you would read the code properly, you have introduced a factory that determines what puzzle should be used. So I think there will be a lot of changes in gameplay-scene to properly clean out those long if conditions since you have this factory. Thoughts?

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -404,17 +411,17 @@ export class GameplayScene {
...
           const newStoneLetter = this.stoneHandler.handleHoveringToAnotherStone(
             trailX,
             trailY,
             (foilStoneText, foilStoneIndex) => {
-              return this.wordPuzzleLogic.handleCheckHoveredStone(foilStoneText, foilStoneIndex);
+              return this.puzzleLogic.handleCheckHoveredStone(foilStoneText, foilStoneIndex);
```
bernhardccodev, line 419: Same here it should be taken a careful look and review the whole scope other than name replacement.

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -505,12 +512,12 @@ export class GameplayScene {
   }
 
   private handleStoneLetterDrawing(deltaTime) {
-    if (this.wordPuzzleLogic.checkIsWordPuzzle()) {
-      const { groupedObj } = this.wordPuzzleLogic.getValues();
+    if (this.puzzleLogic.checkIsWordPuzzle()) {
```
bernhardccodev, line 515: Same here it should be taken a careful look and review the whole scope other than name replacement.

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -40,7 +40,12 @@ import {
   PreviousPlayedLevel,
   MONSTER_PHASES
 } from "@constants";
-import { WordPuzzleLogic } from '@gamepuzzles';
+import { 
+  BasePuzzleLogic, 
+  PuzzleFactory, 
+  FeedbackAudioHandler,
```
bernhardccodev, line 46: Hi @woodsjohn-codev , why do we need both base puzzle logic and puzzle factory?

bernhardccodev (changes requested):
Hi @woodsjohn-codev please check the comments.

ashwinnair-chimple:
@woodsjohn-codev As discussed earlier, I liked the way that you have presented each logic and puzzlefactory in game puzzle but right now as we know we don't have much puzzle other than letter and word so lets simplify it rather than keeping it complex.
by having LetterWord Letter Sound in LetterType class and Word and WordSound in wordType class we can simplify the logic.

PR #1653, opened 2025-04-25
feat: Added Cursor Component and TrailEffectHandler (FM-470).

bernhardccodev:
# Changes
- Cleaned up Trail EFFECT TOGGLE event from GameSettingService, Gameplay-Scene and TrailParticles.
- Created a Cursor Component with ease of adding and removing event methods.
- Created a TrailEffectHandler that will be using the Cursor Component and will be handling trail effects.

# How to test
- Start the FTM app
- Select any game level
- You should be able to see the same trail effect when pressing down.

Ref: [FM-470](https://curiouslearning.atlassian.net/browse/FM-470)

[FM-470]: https://curiouslearning.atlassian.net/browse/FM-470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/cursor/cursor-component.ts
+++ b/src/components/cursor/cursor-component.ts
@@ -1,0 +1,98 @@
+import {
+  CLICK,
+  MOUSEDOWN,
+  MOUSEMOVE,
+  MOUSEUP,
+  TOUCHEND,
+  TOUCHMOVE,
+  TOUCHSTART,
+} from "@common";
+
+export default class CursorComponent {
+  eventHandler: HTMLElement;
+  hasEventsHooked: boolean = false;
+
+  constructor(targetHTML: HTMLElement) {
+    this.eventHandler = targetHTML;
+    // Bind all event handlers
+    this.handleCursorMouseUp = this.handleCursorMouseUp.bind(this);
+    this.handleCursorMouseMove = this.handleCursorMouseMove.bind(this);
+    this.handleCursorMouseDown = this.handleCursorMouseDown.bind(this);
+    this.handleCursorTouchStart = this.handleCursorTouchStart.bind(this);
+    this.handleCursorTouchMove = this.handleCursorTouchMove.bind(this);
+    this.handleCursorTouchEnd = this.handleCursorTouchEnd.bind(this);
+    this.handleCursorMouseClick = this.handleCursorMouseClick.bind(this);
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


src/components/cursor/cursor-component.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.handleCursorTouchMove = this.handleCursorTouchMove.bind(this);
+    this.handleCursorTouchEnd = this.handleCursorTouchEnd.bind(this);
+    this.handleCursorMouseClick = this.handleCursorMouseClick.bind(this);
+  }
+
+  /**
+ * Override this method in a subclass to handle the mouse up event.
+ * This gets triggered when the MOUSEUP event is fired on `eventHandler`.
+ */
+  public handleCursorMouseUp(event) { }
```
ashwinnair-chimple, line 31: remove all the empty methods which are not used, we can keep it if we are gonna handle it with any other related ticket.

bernhardccodev: Hi @ashwinnair-chimple these are all used. If you would check how the Cursor Component is extended, we added the logic on these event methods on the subclass. The point of keeping this empty is to ease the creation of event methods and they are automatically mapped to be added as a listener. This also promotes uniformed but unique instances  for each cursor methods.

src/components/cursor/cursor-component.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+ * Override this method in a subclass to handle the mouse up event.
+ * This gets triggered when the MOUSEMOVE event is fired on `eventHandler`.
+ */
+
+  public handleCursorMouseMove(event) {}
+
+  /**
+ * Override this method in a subclass to handle the mouse up event.
```
ashwinnair-chimple, line 41: first line of all the comments says mouse up!

src/components/cursor/cursor-component.ts
ashwinnair-chimple, line 1: add a destroy method to cursor-component!

bernhardccodev: Hi @ashwinnair-chimple , the cursor has the removeEventListeners method used to remove the add listeners.

ashwinnair-chimple (approved):
approved

PR #1663, opened 2025-05-06
feat: Added RuntimeLoader to use Rive WASM in cached (FM-503)

bernhardccodev:
# Changes
- Cached Rive WASM.
- Added RuntimeLoader to load the Rive WASM in cached instead of relying on Rive CDN to load Rive WASM.

# How to test
- Download the FTM app.
- Go offline and the app should work as it is.

Ref: [FM-503>](https://curiouslearning.atlassian.net/browse/FM-503)


ashwinnair-chimple (approved):
approved

PR #1685, opened 2025-05-20
feat:Fm-533-541 Reinitiating the PR changes in code logic for Hitbox and treelog svg rendering

ashwinnair-chimple:
# Changes
- changes in code logic for Hitbox
-  tree log svg rendering

# How to test
- play the game on multiple devices

Ref: [FM-541](https://curiouslearning.atlassian.net/browse/FM-541)
[FM-540](https://curiouslearning.atlassian.net/browse/FM-540)
[FM-533](https://curiouslearning.atlassian.net/browse/FM-533)

[FM-541]: https://curiouslearning.atlassian.net/browse/FM-541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FM-540]: https://curiouslearning.atlassian.net/browse/FM-540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FM-533]: https://curiouslearning.atlassian.net/browse/FM-533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -226,6 +227,68 @@ export class GameplayScene {
...
+    // Use monster Y position to adjust top if needed
+    const monsterTop = this.monster.getMonsterTopCordinate();
+    if (monsterTop !== undefined && !isNaN(monsterTop)) {
+      svgImg.style.top = `${monsterTop}px`;
+    }
+
+    // Append to parent
+    parent.appendChild(svgImg);
+  }
+
```
bernhardccodev, line 291: Hi @ashwinnair-chimple , this looks great and all. But I think `insertSvgBelowRive` should be a reusable component or at least the logic should separated from gameplay-scene. Thoughts?

src/components/riveMonster/rive-monster-component.ts
```diff
@@ -72,15 +72,49 @@ export class RiveMonsterComponent {
...
+
+    this.hitboxRangeX = {
+      from: (logicalCanvasWidth * 0.5) - (logicalCanvasWidth * 0.3) / 2,
+      to: (logicalCanvasWidth * 0.5) + (logicalCanvasWidth * 0.3) / 2,
+    };
+
+    this.hitboxRangeY = {
+      from: monsterTopY + hitboxPaddingY,
+      to: monsterBottomY - hitboxPaddingY,
+    };
```
bernhardccodev, line 103: Hi @ashwinnair-chimple I think naming and adding more comments to what these computation do would help in the long term maintaining this code.

In the previous code we are using `monsterCenterX` and `monsterCenterY` which indicates the center of screen.

Now we have `logicalCanvasWidth` and `logicalCanvasHeight` along with `monsterBottomY` and `monsterHeight`. I think these names should be updated or add more comments to what these values meant for more context.

Also for the part of `const hitboxPaddingY = monsterHeight * 0.1;` more context would be helpful as to why we need to add padding. Why only 10%? What happens if we use 5%, 20% or more?

Thoughts?

src/components/riveMonster/rive-monster-component.ts
```diff
@@ -256,17 +306,26 @@ The extra space above the monster in the Rive file ensures proper animation, but
           fit: Fit.Contain,
           alignment: Alignment.Center,
         }),
-        onLoad: this.handleLoad.bind(this),
+        onLoad: () => {
+          this.handleLoad();
+          //Recalculate layout and hitbox after phase change
+          this.setRiveMinYAdjustment();
+          this.initializeHitbox();
+        },
```
bernhardccodev, line 314: Hi @ashwinnair-chimple , question. Why do we have so many Rive onload? If this is the method `changePhase`, can you double check if we still need that method? Because if we don't, I don't think it is necessary to keep the method `changePhase` and updating its onLoad. Thoughts?

bernhardccodev (changes requested):
Please see comments.

src/scenes/gameplay-scene/gameplay-scene.ts
ashwinnair-chimple: The main reason why I have not separated it is cause this tree log is only used in gameplay scene and also requires the canvas and hitbox logic, no other place we will use this so gameplay scene seems to be the best place to keep the logic.

src/components/riveMonster/rive-monster-component.ts
ashwinnair-chimple: sure i will add more comments to clear the confusions.

src/components/riveMonster/rive-monster-component.ts
ashwinnair-chimple: so this logic is to update the hitbox when phase 1 to phase 2 or 3 happens the height of the monster changes thus updating the hitbox range and artboard, we can also call it on the onload method but as changephase is the first method that gets triggered i have updated it here. Let me know if you have any other ideas. @bernhardccodev

bernhardccodev (approved):

woodsjohn-codev (approved):

PR #1648, opened 2025-04-22
feat: Added Letter Puzzle Tutorial (FM-482)

bernhardccodev:
# Changes
- Deleted the old Tutorial class.
- Added Letter Puzzle Tutorial classed using the Base Tutorial from FM-481
- Update Base Tutorial method for stone letter guide.
- Update gameplay-scene by replacing the old tutorial component to Letter Puzzle Tutorial.
- Cleaned up stone-handler and stone-config by removing tutorial.
- Updated gameStateService by adding another event tailored to connect stone-handler stone creation details to Letter Puzzle.
- Created test case for Letter Puzzle 

# How to test
- Run FTM APP
- Select and play game level 1.
- In the start there should be a pointer indicating and guiding to the quick start feature.
- As soon as the letter drops there should be a guide to drag the correct letter to the hitbox.

Ref: [FM-482](https://curiouslearning.atlassian.net/browse/FM-482)

[FM-482]: https://curiouslearning.atlassian.net/browse/FM-482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1686, opened 2025-05-21
feat: Created Tutorial Handler for handling Tutorials (FM-542)

bernhardccodev:
# Changes
- Created Tutorial Handler that will handler the different tutorial initialization and logic to when to draw and hide tutorials.
- Modified MatchLetterPuzzle by moving the unnecessary logic that should be for Tutorial Handler.
- Separated QuickStartTutorial so it can easily be managed.
- Clean up BaseTutorialComponent as it has the logic that should be for Tutorial-Handler.
- Updated GameStateService by having the logic to determine the first instances of each game type puzzles to determine which levels should have Tutorial.
- Updated test case and created new one for Tutorial-Handler.

# How to test
- Start FTM app.
- Since we only have one tutorial for now, it should show the first Match Letter Tutorial in game level 1, segment 1 only,

Ref: [FM-542](https://curiouslearning.atlassian.net/browse/FM-542)

[FM-542]: https://curiouslearning.atlassian.net/browse/FM-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/gameStateService/gameStateService.ts
```diff
@@ -118,24 +126,74 @@ export class GameStateService extends PubSub {
...
         this.monsterPhaseNumber = this.checkMonsterPhaseUpdation();
+        this.gameTypesFirstInstanceList = this.getFirstInstanceOfEachGameTypes(data);
+    }
+
+    private getFirstInstanceOfEachGameTypes(data: DataModal) {
+        if (data.levels) {
+            //Determine the first time game types will appear.
+            const levelList: any = data.levels;
+            const gameTypes = {};
+            let levelType = null;
```
genai-chimpledev, line 137: instead of calling  levelType, levelNumber, protoType outside the loop you can use inline 
const { levelType, levelNumber, protoType } = levelData?.levelMeta ?? {};

ashwinnair-chimple, line 137: instead of calling  levelType, levelNumber, protoType outside the loop.
this can be declared in line as 
const { levelType, levelNumber, protoType } = levelData?.levelMeta ?? {};

src/tutorials/index.ts
```diff
@@ -1,3 +1,143 @@
...
+            * stoneDataImgAndPos contains the stone details from stone handler.
+            * It has the correct stone coordinates needed for the tutorial of dragging stones.
+            */
+            const { stonePosVal, img, levelData } = stoneDataImgAndPos;
+            const { protoType, levelType } = levelData?.levelMeta;
+            const levelNumber = levelData?.levelNumber;
+            const gameLevel = typeof levelNumber === 'string'
+              ? parseInt(levelNumber)
+              : levelNumber;
+            const gameTypeName = protoType === 'Hidden' ? `Sound${levelType}` : levelType
```
ashwinnair-chimple, line 63: @bernhardccodev I think we can move gameTypeName into Utils as I feel this will be reusable. Thoughts?

src/tutorials/index.ts
```diff
@@ -1,3 +1,143 @@
...
+      this.activeTutorial = null;
+      this.quickTutorial = null;
+    }
+  }
+
+  drawQuickStart(deltaTime: number, hasGameStarted: boolean) {
+    if (!hasGameStarted && this.quickTutorial) {
+      //Show quick tip by pressing the center/near monster.
+      this.quickTutorial.quickStartTutorial(deltaTime, this.width, this.height);
+    }
```
ashwinnair-chimple, line 122: how about combining  drawQuickStart() and draw() method which pass the same paramaters

src/tutorials/index.ts
```diff
@@ -1,3 +1,143 @@
...
+
+  dispose() {
+    //Clear canvas tutorials and reset values;
+    if (this.hasEstablishedSubscriptions) {
+      this.isGameOnPause = false;
+      this.hasGameEnded = false;
+      this.quickTutorial = null;
+      this.activeTutorial = null;
+      this.unsubscribeStoneCreationEvent();
+      this.unsubscribePauseEvent();
```
ashwinnair-chimple, line 140: Thoughts? @bernhardccodev 
can we set this.hasEstablishedSubscriptions = false  in dispose to reuse it ?

src/tutorials/index.ts
ashwinnair-chimple: i was just thinking method remains one and logic we can manage inside it!!

src/tutorials/index.ts
bernhardccodev: Hi @ashwinnair-chimple , I don't this will be reusable. The gameTypeName is just to access the gameTypes created by the gameStateService which contains the list of first instances so we can tell when to display tutorials and its corrisponding tutorial levels.

Since this is mainly use for tutorial, it won't make sense to move it for a utils where only one file would benefit.

src/gameStateService/gameStateService.ts
```diff
@@ -118,24 +126,74 @@ export class GameStateService extends PubSub {
...
             : "";
 
+        let shouldHaveTutorial = false;
+        const selectedLevelNumber:string | number = this.gamePlayData.selectedLevelNumber;
+        const levelNumber = typeof selectedLevelNumber === 'string' ? parseInt(selectedLevelNumber) : selectedLevelNumber;
+        //Very small array to iterate.
+        Object.values(this.gameTypesFirstInstanceList).forEach((listedLevelNumber) => {
+            if (listedLevelNumber === levelNumber ) {
+                shouldHaveTutorial = true;
+                return; //stop the loop;
```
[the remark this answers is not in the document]
bernhardccodev: Hi @ashwinnair-chimple, The comment indicates its just to stop the loop. We don't need to return anything here, we just need to validate if the level number AKA game level selected is on the list to have a tutorial.

src/gameStateService/gameStateService.ts
bernhardccodev: Hi @ashwinnair-chimple , yeah that make sense. Let me apply that.

src/tutorials/index.ts
bernhardccodev: Hi @ashwinnair-chimple I'll apply that.

src/tutorials/index.ts
bernhardccodev: Hi @ashwinnair-chimple, below here is the gameplay-scene draw method.
You can see here there is a return if the game doesn't start yet. The reason the quickTutorial needs to be separated is that it will be drawn before the game starts. otherwise it wont be drawn.

`draw(deltaTime: number) {
    if (!this.isGameStarted && !this.isPauseButtonClicked) {
      this.time += deltaTime;
      this.tutorial.drawQuickStart(deltaTime, this.isGameStarted);
      if (this.time >= 5000) {
        this.setGameToStart();
      }
      // Don't draw game elements until started
      return;
    }

    this.trailEffectHandler?.draw();

    if (this.isGameStarted) {
      this.handleStoneLetterDrawing(deltaTime);
    }

    this.tutorial.draw(deltaTime, this.isGameStarted);
  }`

src/tutorials/index.ts
```diff
@@ -1,3 +1,144 @@
...
+  */
+  public hideTutorial() {
+    if (this.activeTutorial) {
+      this.puzzleLevel++;
+      this.activeTutorial = null;
+      this.quickTutorial = null;
+    }
+  }
+
+  drawQuickStart(deltaTime: number, hasGameStarted: boolean) {
```
woodsjohn-codev, line 118: can we add a comment for this method?

bernhardccodev: Hi @woodsjohn-codev the name itself is self explanatory.

src/tutorials/index.ts
```diff
@@ -1,3 +1,144 @@
...
+    this.width = width;
+    this.height = height;
+    this.context = context;
+    this.quickTutorial = new QuickStartTutorial({ context: this.context });
+  }
+
+  private createTutorialInstance({ gameLevel, stonePosVal, img, gameTypeName }: {
+    gameLevel: number,
+    stonePosVal: number[],
+    img: any,
```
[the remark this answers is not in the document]
bernhardccodev: Hi @woodsjohn-codev , no its not an HTMLImageElement.

You can refer to this PR for reference: https://github.com/curiouslearning/FeedTheMonsterJS/pull/1648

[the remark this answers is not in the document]
woodsjohn-codev: yeah i see it. thanks Berns @bernhardccodev !

woodsjohn-codev (approved):
Looks good to me

ashwinnair-chimple (approved):
approved

PR #1700, opened 2025-05-30
fix: Allowed retry scene name to be re-instance (FM-554).

bernhardccodev:
# Changes
- Allowed Replay Scene Flag to be reinitialized even if the scene name is already Replay-Scene.

# How to test
- Run FTM App
- Select any game level
- At the game start perform the retry button.
- As the game starts again, retry for the second time.
- App should still reply the game level.

Ref: [FM-554](https://curiouslearning.atlassian.net/browse/FM-554)

### DIFF
--- a/src/sceneHandler/scene-handler.ts
+++ b/src/sceneHandler/scene-handler.ts
@@ -50,7 +50,11 @@ export class SceneHandler {
     this.unsubscribeEvent = gameStateService.subscribe(
       gameStateService.EVENTS.SWITCH_SCENE_EVENT,
       (sceneName: string) => {
-        if (!this.currentScene || this.currentScene !== sceneName) {
+        if (
+          !this.currentScene ||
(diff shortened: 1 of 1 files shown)


src/sceneHandler/scene-handler.ts
```diff
@@ -50,7 +50,11 @@ export class SceneHandler {
     this.unsubscribeEvent = gameStateService.subscribe(
       gameStateService.EVENTS.SWITCH_SCENE_EVENT,
       (sceneName: string) => {
-        if (!this.currentScene || this.currentScene !== sceneName) {
+        if (
+          !this.currentScene ||
+          this.currentScene !== sceneName || //Prevents the double instance of the same scene (fix from FM-532 issue).
```
ashwinnair-chimple, line 55: @bernhardccodev  The logic looks good! Just a small suggestion,
when we have conditions that can be a bit hard to read at first glance, it might help to extract them into a const for better readability and understanding. Makes the intent clearer for everyone reviewing the code.

PR #1713, opened 2025-06-09
feat: FM-544 Start the timer only after the tutorial interaction or after duration.

ashwinnair-chimple:
# Changes
- Added duration of 6s to QuickStart tutorial
- Added duration of 12s to Drag Tutorial
- If user interacts between Drag tutorial then start the timer ticking
- If user does not  interacts between Drag tutorial then start the timer ticking after 12s

# How to test
-  Play the 1st level of match first as a new player

Ref: [FM-544](https://curiouslearning.atlassian.net/browse/FM-544)

[FM-544]: https://curiouslearning.atlassian.net/browse/FM-544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -514,9 +518,14 @@ export class GameplayScene {
     if (this.stoneHandler.stonesHasLoaded && !this.isPauseButtonClicked) {
       if (this.shouldShowTutorialAnimation) {
         // FM-544 add or modify code logic here to controlling the timer when tutorial is animating.
-        this.timerTicking.update(deltaTime);
+        this.tutorialElapsedTime += deltaTime;
+        if (this.tutorialElapsedTime >= this.tutorialHoldDuration) {
```
bernhardccodev, line 522: Hi @ashwinnair-chimple , PR looks great. Just minor suggestion, and your inputs is needed here. What do you think if we move :
`private tutorialElapsedTime: number = 0;
  private readonly tutorialHoldDuration: number = 12000; // 12 seconds`
  
  to tutorial-handler.ts, have it as:
`public tutorialElapsedTime: number = 0;
  public readonly tutorialHoldDuration: number = 12000; // 12 seconds`

then in `draw` method of tutorial-handler.ts
insert this part `this.tutorialElapsedTime += deltaTime;`

So all logic that is related to tutorial will be in tutorial-handler. We can just acceess the `tutorialElapsedTime` and `tutorialHoldDuration` result with a new method something like isTutorialDurationOver:

`isTutorialDurationOver(){
 return this.tutorialElapsedTime >= this.tutorialHoldDuration
}`

In this way, gameplay-scene would careless about the tutorial logic and the relating tutorial logic will be handled in the tutorial class like the rest of the logic for tutorial. Thoughts?

ashwinnair-chimple: thanks for this suggestion, this sounds good!  I will implement this @bernhardccodev

bernhardccodev (approved):
Looks good bro!

woodsjohn-codev (approved):
Looks good to me

PR #1728, opened 2025-06-18
feat: AJ-338, FM-562 analytics integration

dz4va:
# Changes
- Added `@cl/analytics` package for use in the Firebase/
- Modified the `FirebaseIntegration` `initialization` and `getInstance` methods and separated them, because of the fact that the analytics package is using an async way of initialization
- Added the `FirebaseIntegration` initialization logic in the `feedTheMonster.ts` async `init` method
- Made sure that we are using `FirebaseIntegration.getInstance()` everywhere in the codebase and not creating new objects every single time
- Fixed Firebase analytics mocking in test files by replacing class-based mocks with Jest function mocks
- Updated the tests in `firebase-integration.spec.ts`
- Updated `gameplay-scene.spec.ts` to use proper mock implementation with fresh instances per test
- Updated `start-scene.spec.ts` to align with the new analytics package mocking pattern
- Removed redundant `resetInstance` calls in favor of fresh mock instances through `initializeAnalytics`
- Ensured mock implementations match the actual FirebaseIntegration interface from `@cl/analytics`

# How to test
- Run `npm test` and verify all tests in `firebase-integration.spec.ts`, `gameplay-scene.spec.ts` and `start-scene.spec.ts` pass without any errors
- In the testing apk and later on in BigQuery and Statsig to see if we are getting the actual user data

Ref: [AJ-338](https://curiouslearning.atlassian.net/browse/AJ-338) , [FM-562](https://curiouslearning.atlassian.net/browse/FM-562)

[AJ-338]: https://curiouslearning.atlassian.net/browse/AJ-338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Firebase/base-firebase-integration.ts
```diff
@@ -1,49 +1,99 @@
...
+
+    get firebaseApp() {
+        return this.firebaseStrategy?.firebaseApp;
+    }
+
+    // Method to check if analytics is ready
+    public isAnalyticsReady(): boolean {
+        return this.isInitialized;
     }
 }
```
ashwinnair-chimple, line 99: Hey @dz4va , since FTM initializes its own analytics (Firebase + Statsig), do you think it makes sense to add a dispose() method in the BaseFirebaseIntegration class?

That way, if the user switches to another sub-app, we can cleanly shut down Statsig (shutdown()) and optionally delete the Firebase app instance to avoid lingering memory or event issues. Might help prevent conflicts if other sub-apps also use analytics.

Curious to hear what you think!

PR #1745, opened 2025-07-01
fix: Star Count Increases on Level Failures, skipping Monster Evolution Animation (FM-580)

woodsjohn-codev:
# Changes
- fix: Prevent 0-1 Star Levels from Contributing to Evolution Progress
- Modified the star count calculation logic to only include levels with 2 or more stars in the total star count used for evolution mechanics, while still displaying the correct visual representation of stars (0-3) in the UI.

# How to test
- Play level 1: Earn 2 stars
- Play level 2: Earn 2 stars
- Play level 3: Earn 3 stars
- Total accumulated: 7 stars (all counting toward evolution)
- Play level 4: Deliberately earn only 0 or 1 star
- Verify the star is displayed visually in the UI
- Verify the evolution animation does NOT trigger

Ref: [FM-580](https://curiouslearning.atlassian.net/browse/FM-580)

[FM-580]: https://curiouslearning.atlassian.net/browse/FM-580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/data/game-score.ts
+++ b/src/data/game-score.ts
@@ -42,7 +42,8 @@ export class GameScore {
   private static updateTotalStarCount(): void {
     const allGameLevelInfo = this.getAllGameLevelInfo();
     const totalStarCount = allGameLevelInfo.reduce(
-      (sum, level) => sum + level.starCount,
+      // Only count levels with 2 or more stars toward the total
+      (sum, level) => sum + (level.starCount >= 2 ? level.starCount : 0),
       0
     );
     localStorage.setItem(this.currentlanguage + "totalStarCount", totalStarCount.toString());
--- a/src/data/game-score.ts
+++ b/src/data/game-score.ts
@@ -43,7 +43,7 @@ export class GameScore {
     const allGameLevelInfo = this.getAllGameLevelInfo();
     const totalStarCount = allGameLevelInfo.reduce(
       // Only count levels with 2 or more stars toward the total
-      (sum, level) => sum + (level.starCount >= 2 ? level.starCount : 0),
+      (sum, level) => sum + level.starCount,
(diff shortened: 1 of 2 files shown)


src/data/game-score.ts
```diff
@@ -42,7 +42,8 @@ export class GameScore {
   private static updateTotalStarCount(): void {
     const allGameLevelInfo = this.getAllGameLevelInfo();
     const totalStarCount = allGameLevelInfo.reduce(
-      (sum, level) => sum + level.starCount,
+      // Only count levels with 2 or more stars toward the total
+      (sum, level) => sum + (level.starCount >= 2 ? level.starCount : 0),
```
bernhardccodev, line 46: Hi @woodsjohn-codev,
I'm a bit confused about this change — how is this only affecting the monster evolution logic when you're modifying the actual total star count being saved to local storage?

While I understand we're only considering levels with 2 or more stars for the monster evolution, this logic seems specific to that feature. Updating the total star count here feels risky since it affects the global state used across the app. The total star count stored should reflect all stars earned per level (for display and other features), not just those applicable to the monster evolution criteria.

I suggest isolating the logic for monster evolution instead of modifying the app-wide total. Let me know your thoughts.

bernhardccodev (changes requested):
Hi @woodsjohn-codev , Please revisit this — it feels like a shortcut that affects a global value just to support a feature-specific case, and that could introduce unintended issues. Thanks!

ashwinnair-chimple (approved):
approved

PR #1773, opened 2025-07-30
fix: FM-568 The drag tutorial gets interrupted when device screen goes to sleep while the tutorial is playing. 

ashwinnair-chimple:
# Changes
- Code enhancement

# How to test
- The drag tutorial gets interrupted when device screen goes to sleep while the tutorial is playing. 

Ref: [FM-568](https://curiouslearning.atlassian.net/browse/FM-568)

[FM-568]: https://curiouslearning.atlassian.net/browse/FM-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/tutorials/base-tutorial/base-tutorial-component.ts
+++ b/src/tutorials/base-tutorial/base-tutorial-component.ts
@@ -31,6 +31,9 @@ export default class TutorialComponent {
   public x: number = 0;
   public y: number = 0;
   public totalTime: number = 0;
+  public frame: number = 0;
+  protected animationStartTime: number = 0;
+  protected maxDeltaTime: number = 33; // Cap deltaTime to 33ms (approx 30fps) for consistent animation
   private centerX: number = 0;
   private centerY: number = 0;
   private initialOuterRadius: number = 10
@@ -146,7 +149,7 @@ export default class TutorialComponent {
         from: number,
         to: number
       }
-  } = gameStateService.getHitBoxRanges();
+    } = gameStateService.getHitBoxRanges();
     const getRangeCenter = (hitBoxFromPos: number, hitBoxToPos: number) => {
       return (hitBoxFromPos + hitBoxToPos) / 2;
     }
@@ -236,7 +239,7 @@ export default class TutorialComponent {
       this.context.drawImage(this.tutorialImg, this.x + 15, this.y + 10);//draws the hand stone drag animation!
     }
   }
-  
+
   /**
    * Specialized animation for word puzzle tutorials
    * Provides a more guided animation path with visual cues for multi-letter words
@@ -264,7 +267,7 @@ export default class TutorialComponent {
   }) {
     // Draw the stone at current position with slightly higher opacity for word puzzles
     const previousAlpha = this.context.globalAlpha;
-  
+
     // Near the start position
     if (monsterStoneDifferenceInPercentage > 80) {
       this.context.globalAlpha = 0.8;
@@ -278,10 +281,10 @@ export default class TutorialComponent {
         // Draw at the CURRENT position instead of a fixed position
         // This ensures continuous movement all the way to the end
         this.context.drawImage(img, this.x, this.y, imageSize, imageSize);
-        
+
         // Move the hand with the stone
         this.createHandScaleAnimation(deltaTime, this.x + 15, this.y + 10, true);
-        
+
         // Add a subtle highlight effect at the destination
         this.context.beginPath();
         this.context.arc(endX, endY, imageSize/2, 0, Math.PI * 2);
@@ -299,7 +302,7 @@ export default class TutorialComponent {
       this.context.globalAlpha = previousAlpha;
       this.context.drawImage(this.tutorialImg, this.x + 15, this.y + 10);
     }
-    
+
     this.context.globalAlpha = previousAlpha;
   }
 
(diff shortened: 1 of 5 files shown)


src/tutorials/base-tutorial/base-tutorial-component.ts
```diff
@@ -319,6 +323,123 @@ export default class TutorialComponent {
...
+   * @param maxFrame The maximum frame value to reach
+   * @param animationDuration The duration of the animation in milliseconds
+   */
+  protected updateAnimationFrame(maxFrame: number, animationDuration: number): void {
+    if (this.frame < maxFrame) {
+      if (this.animationStartTime === 0) {
+        this.animationStartTime = performance.now();
+      }
+      const elapsed = performance.now() - this.animationStartTime;
+      this.frame = Math.min(maxFrame, (elapsed / animationDuration) * 100);
```
bernhardccodev, line 337: Noticing something potentially fragile here:

The line:
`this.frame = Math.min(maxFrame, (elapsed / animationDuration) * 100);`

hardcodes a cap at 100 while also using a variable maxFrame. This works now because both values happen to be 100, but:

- If` maxFrame > 100`, the frame will never exceed 100, regardless of intent.

- If `maxFrame < 100`, the animation finishes early since it hits `maxFrame `before `elapsed `reaches `animationDuration`.

So we have a hardcoded normalization scale (`* 100`) that could easily drift from `maxFrame`, even though `maxFrame `implies flexibility. This creates a misleading expectation that you can tune `maxFrame `freely, when in practice it's tightly coupled to that `100`.

TL;DR: There's a risk here of unintentionally breaking animation timing if maxFrame is ever changed without adjusting this `* 100`.

src/tutorials/base-tutorial/base-tutorial-component.ts
```diff
@@ -319,6 +323,123 @@ export default class TutorialComponent {
...
 
+  /**
+   * Updates the animation frame based on elapsed time
+   * @param maxFrame The maximum frame value to reach
+   * @param animationDuration The duration of the animation in milliseconds
+   */
+  protected updateAnimationFrame(maxFrame: number, animationDuration: number): void {
+    if (this.frame < maxFrame) {
+      if (this.animationStartTime === 0) {
+        this.animationStartTime = performance.now();
```
bernhardccodev, line 334: This condition only passes once, since `animationStartTime` is never reset to `0` afterward. Might be worth clarifying that this acts as a one-time initializer. Otherwise, it can look like a recurring check at first glance — when in reality, it's a one-shot setup that won't re-run unless explicitly reset elsewhere.

src/tutorials/base-tutorial/base-tutorial-component.ts
```diff
@@ -319,6 +323,123 @@ export default class TutorialComponent {
...
+    // Calculate current and next distance percentages
+    const current = this.calculateDistancePercentage(this.x, this.y);
+    const next = this.calculateDistancePercentage(nextX, nextY);
+
+    // Critical animation range (1-15%): ensure we don't skip it
+    const CRITICAL_PERCENTAGE = 14;
+    if (current.percentage > 15 && next.percentage < 15) {
+      // Adjust position to ensure we hit the critical percentage mark
+      this.adjustPositionToPercentage(current.percentage, CRITICAL_PERCENTAGE, current.disX, current.disY);
+    } else {
```
bernhardccodev, line 433: Hi @ashwinnair-chimple , can you clarify the purpose of `CRITICAL_PERCENTAGE` here? I see it’s hardcoded to `14`, and the logic adjusts the position if we're about to skip over it. Just trying to understand what behavior this is meant to protect — is there a visual or logic trigger around that percentage range?

bernhardccodev (changes requested):
Hi @ashwinnair-chimple,

Thanks for the changes — I’ve left a few comments/questions above, mostly around clarity of the logic and potential edge cases.

I also checked the JIRA ticket linked to this PR, but there wasn’t much detail on how the bug actually occurs. Without that context, I’m having a hard time reviewing this code with confidence that it addresses the underlying issue — especially since it affects tutorial rendering logic.

If you could share more background on how the bug manifests and how these changes are meant to solve it, that would really help in evaluating the fix properly.

src/tutorials/base-tutorial/base-tutorial-component.ts
ashwinnair-chimple: Hey @Bernhard Cena  great question!

Yes, the CRITICAL_PERCENTAGE is guarding a visual behavior. Around the 14–15% mark, the stone overlaps closely with the monster’s starting position. On slower devices, the animation can skip over that zone too quickly, causing the overlap moment to be missed which visually looks like the stone "jumps" forward.

This logic ensures we don't miss that overlap phase, which is important for syncing players guidance. It acts as a visual safeguard rather than a logic trigger.

If required we can rename it.

bernhardccodev (approved):
Hi @ashwinnair-chimple, thanks for the clarification on `CRITICAL_PERCENTAGE`. That really helped in understanding its purpose, and if that’s the case, then I’m aligned with the approach. Approving this now.

PR #1781, opened 2025-08-13
enhancement: Updated prompt bubble rendering and its CSS styling (FM-597)

bernhardccodev:
# Changes
- Updated Prompt Bubble rendering by moving it from background-image style of a dive into img html tag.
- Updated the CSS styling to accommodate the new img tag for prompt bubble.

# How to test
- Run FTM app.
- Play any game level and expect prompt bubble should be properly aligned with the audio button or text prompt.

Ref: [FM-597](https://curiouslearning.atlassian.net/browse/FM-597)

[FM-597]: https://curiouslearning.atlassian.net/browse/FM-597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/prompt-text/prompt-text.scss
+++ b/src/components/prompt-text/prompt-text.scss
@@ -18,7 +18,7 @@
   }
 }
 
-.responsive-position {
+.prompt-center-responsive {
   position: relative;
   left: 50%;
   transform: translateX(-50%);
@@ -58,7 +58,7 @@
 }
 
 //Custom position and size for prompt bubble. Used for Spell Word Audio puzzle.
-.prompt-bubble-custom {
+.prompt-bubble-spell-audio {
 
   // Small mobile (≤376px)
   @media (max-width: 376px) {
@@ -70,7 +70,7 @@
   // Large mobile (377-480px)
   @media (min-width: 377px) and (max-width: 480px) {
     width: 280px;
-    height: 240px;
+    height: 245px;
     top: 13%;
   }
 
--- a/src/components/prompt-text/prompt-text.scss
+++ b/src/components/prompt-text/prompt-text.scss
@@ -18,14 +18,8 @@
   }
 }
 
-
-
-.prompt-background {
+.responsive-position {
   position: relative;
-  background-size: contain;
-  background-repeat: no-repeat;
-  background-position: center;
-  transition: transform 0.05s ease;
   left: 50%;
   transform: translateX(-50%);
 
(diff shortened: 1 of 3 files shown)


src/components/prompt-text/prompt-text.scss
```diff
@@ -18,14 +18,8 @@
   }
 }
 
-
-
-.prompt-background {
+.responsive-position {
```
ashwinnair-chimple, line 21: Hey @bernhardccodev 
can we change the naming to something related to the element?

src/components/prompt-text/prompt-text.scss
```diff
@@ -58,36 +52,40 @@
...
-// TEMP: Spell Sound only - overrides prompt bubble size
-// This is a band-aid fix to force specific dimensions for that gameplay.
-// Consider refactoring layout structure if background scaling becomes shared.
+#prompt-bubble {
+  position: absolute;
+  transition: transform 0.05s ease;
+}
+
+//Custom position and size for prompt bubble. Used for Spell Word Audio puzzle.
 .prompt-bubble-custom {
```
ashwinnair-chimple, line 61: I think so we should define the media query at a common level rather than nesting it inside specific selectors.

src/components/prompt-text/prompt-text.ts
```diff
@@ -10,7 +10,7 @@ import gameStateService from '@gameStateService';
 export const DEFAULT_SELECTORS = {
     root: '#background', // The background element to append the prompt to
     promptText: '.prompt-text',
-    promptBackground: '.prompt-background',
+    promptBackground: '.prompt-position',
```
ashwinnair-chimple, line 13: can we change the naming to something related to the element as mentioned above?

src/components/prompt-text/prompt-text.ts
```diff
@@ -32,10 +32,15 @@ export const PROMPT_TEXT_LAYOUT = (
 
     return (`
         <div id="${id}" class="prompt-container ">
+            <img
+                id="prompt-bubble"
```
ashwinnair-chimple, line 36: Hey @bernhardccodev ,
I think having a parent image container will give us more flexibility to apply CSS, rather than applying styles directly to the image.
We can set the container size as needed, and the image can then adapt using max-width: 100%.
your thoughts?

src/components/prompt-text/prompt-text.ts
```diff
@@ -32,10 +32,15 @@ export const PROMPT_TEXT_LAYOUT = (
...
+                class="responsive-position ${bubblePulsateStyle}"
+                src="${PROMPT_TEXT_BG}"
+                alt="audio button"
+            />
             <div
-                id="prompt-background"
-                class="prompt-background ${bubblePulsateStyle} "
-                style="background-image: url(${PROMPT_TEXT_BG})"
+                id="prompt-content"
+                class="responsive-position ${bubblePulsateStyle}"
```
ashwinnair-chimple, line 43: can we change the naming to something related to the element as mentioned above?

bernhardccodev: Hi @ashwinnair-chimple ,

Changes applied - changed `responsive-position` to `prompt-center-responsive`

src/components/prompt-text/prompt-text.scss
bernhardccodev: Hi @ashwinnair-chimple ,

I renamed this from .prompt-bubble-custom to .prompt-bubble-spell-audio.
This style only applies to Spell Sound puzzles, where the prompt bubble needs a specific size and position to fit the audio button and slot texts.

Since it’s not used in most gameplays, keeping the media queries scoped here keeps it self-contained and avoids adding unused rules at a global level.

ashwinnair-chimple (approved):
looks good approved!

PR #1807, opened 2025-09-15
feat: Trigger treasure chest mini game mechanics implementation (FM-594)

bernhardccodev:
# Changes
- Created logic that determines the list of game levels to where mini game would show up.
- Updated logic in triggering mini games by adding the level checking from the list of game levels for mini game along side with randomizer for game level, level segments. Now checks if mini game is cleared.
- Updated Game Score logic to save mini game score.
- Code clean up and updated test case.

# How to test
- Run FTM app.
- On game levels 2, 5, 15, 25.. the mini game should trigger.
- Mini game will appear randomly within the game level segment.
- Scores should be visible in the local app storage.

Ref: [FM-594](https://curiouslearning.atlassian.net/browse/FM-594)

### DIFF
--- a/src/data/game-score.ts
+++ b/src/data/game-score.ts
@@ -3,25 +3,32 @@ import { Debugger, lang } from "@common";
 export class GameScore {
   public static currentlanguage: string = lang;
 
-  public static setGameLevelScore(currentLevelInfo, score) {
+  public static setGameLevelScore(currentLevelInfo, score, treasureChestMiniGameScore) {
     let starsGained = this.calculateStarCount(score);
     let levelPlayedInfo = {
       levelName: currentLevelInfo.levelMeta.levelType,
       levelNumber: currentLevelInfo.levelMeta.levelNumber,
       score: score,
       starCount: starsGained,
+      treasureChestMiniGameScore,
     };
     let allGameLevelInfo = this.getAllGameLevelInfo();
     let index = allGameLevelInfo.findIndex(
       (level) => level.levelNumber === levelPlayedInfo.levelNumber
     );
 
+    //If there are data found in local storage, cross check the scores.
     if (index !== -1) {
-      // Update only if the new score is higher
-      if (levelPlayedInfo.score > allGameLevelInfo[index].score) {
+      //Check if the new scores are higher.
+      const isNewScoreHigher = score > allGameLevelInfo[index].score;
+      const isNewMiniGameScoreHigher = treasureChestMiniGameScore > allGameLevelInfo[index].treasureChestMiniGameScore;
+
+      // Update only if the new score is higher either from mini game or actual game score.
+      if (isNewScoreHigher || isNewMiniGameScoreHigher) {
         allGameLevelInfo[index] = levelPlayedInfo;
       }
     } else {
+      // If the game level is newly cleared.
       allGameLevelInfo.push(levelPlayedInfo);
     }
 
--- a/src/miniGame/miniGameStateService/miniGameStateService.ts
+++ b/src/miniGame/miniGameStateService/miniGameStateService.ts
@@ -1,9 +1,15 @@
 import { PubSub } from '../../events/pub-sub-events';
+import { GameScore } from "@data";
 
+type MiniGameMap = {
+  [key: number]: { isMiniGameComplete: boolean };
+};
 export class MiniGameStateService extends PubSub {
   public EVENTS: {
     IS_MINI_GAME_DONE: string;
   }
+
+  private treasureChestCompletedLevel: MiniGameMap;
 
   constructor() {
     super();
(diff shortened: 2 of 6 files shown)


src/data/game-score.ts
```diff
@@ -3,25 +3,32 @@ import { Debugger, lang } from "@common";
 export class GameScore {
   public static currentlanguage: string = lang;
 
-  public static setGameLevelScore(currentLevelInfo, score) {
+  public static setGameLevelScore(currentLevelInfo, score, treasureChestMiniGameScore) {
     let starsGained = this.calculateStarCount(score);
     let levelPlayedInfo = {
```
janfb-codev, line 8: We should probably use a type or interface for this struct since from the looks of it, it was used on different parts of the codebase as well.

src/data/game-score.ts
```diff
@@ -3,25 +3,32 @@ import { Debugger, lang } from "@common";
...
+    //If there are data found in local storage, cross check the scores.
     if (index !== -1) {
-      // Update only if the new score is higher
-      if (levelPlayedInfo.score > allGameLevelInfo[index].score) {
+      //Check if the new scores are higher.
+      const isNewScoreHigher = score > allGameLevelInfo[index].score;
+      const isNewMiniGameScoreHigher = treasureChestMiniGameScore > allGameLevelInfo[index].treasureChestMiniGameScore;
+
+      // Update only if the new score is higher either from mini game or actual game score.
+      if (isNewScoreHigher || isNewMiniGameScoreHigher) {
```
janfb-codev, line 27: This could result an issue where the minigame score is higher but the score is lower. Not sure with the score calculation with this if it is affected.

src/miniGame/miniGameStateService/miniGameStateService.ts
```diff
@@ -1,26 +1,97 @@
...
+    this.treasureChestCompletedLevel = completedLevel;
   }
 
-  selectLevelAtRandom(levelSegmentLength: number) {
+  private selectLevelAtRandom(levelSegmentLength: number) {
     return Math.floor(Math.random() * levelSegmentLength) + 1;
   }
-}
 
+  public shouldShowMiniGame({
```
janfb-codev, line 81: What is the return type for this function? It feels like it is a bool function but it returns a number

src/miniGame/miniGameStateService/miniGameStateService.ts
```diff
@@ -1,26 +1,97 @@
...
+      this.treasureChestCompletedLevel[adjustGameLevelValue].isMiniGameComplete = miniGameScore === 1;
+    });
+  }
+
+  //Creates levels of completed list of treasure chest minigame.
+  private initMiniGameLevelList() {
+    const completedLevel: MiniGameMap = {
+      2: { isMiniGameComplete: false }, //starting level for mini game as per FM-594; not a special level.
+    }
+    let startingSpecialLevel = 5;
```
ashwinnair-chimple, line 43: Hey @bernhardccodev, The starting special level (5), special level interval (10) and maximum level (200) are hardcoded. Consider extracting these as named constants for better maintainability.

src/miniGame/miniGameStateService/miniGameStateService.ts
```diff
@@ -1,26 +1,97 @@
...
   }
 
   private initListeners() {
     /* Listeners to update mini game state values or trigger certain logics. */
+    this.subscribe(this.EVENTS.IS_MINI_GAME_DONE, ({ miniGameScore, gameLevel }: {
+      miniGameScore: number,
+      gameLevel: number,
+    }) => {
+      //Plus one to match the expected level as game level starts at 0 whereas business logic starts at 1.
+      const adjustGameLevelValue = gameLevel + 1;
```
ashwinnair-chimple, line 32: The game level adjustment (+1) is repeated in multiple places. Consider extracting this into a helper method to avoid duplication and potential inconsistencies.
@bernhardccodev

bernhardccodev: Hi @ashwinnair-chimple , It only happens in two classes: gameplay-scene and miniGameStateService. I don’t think this justifies creating a helper for only these cases.

The +1 is simply to align this.counter with gamePlayData.levelData.levelNumber. Since arrays start at index 0, we need to shift by 1 to match our expected list of levels where the mini game can appear (which starts at 1). This is more of a business logic decision, and because we don’t have proper documentation, I made it start at 1 rather than 0 for better readability.

Not all classes will ever need the +1. Only miniGameStateService and gameplay-scene require it, since they’re the ones that need to determine which level will trigger the mini game based on the random level number provided by miniGameStateService.

ashwinnair-chimple: Thanks for clarifying, @bernhardccodev 
I understand your point that the +1 is tied to business logic and only shows up in those two classes. My main thought was around future maintainability if this logic ever needs to change, we’d have to track it in multiple places.

That said, I’m fine leaving it as-is for now since it’s isolated to gameplay-scene and miniGameStateService. Maybe we can just drop a short comment in those spots explaining the reasoning behind the +1 (array index vs. level numbering) so it’s clearer for anyone reading the code later.

Would that work?

bernhardccodev: Thanks @ashwinnair-chimple , I did also put a thought on that. But I just realize not every class will be interacting with the mini game level list so the +1 is manageable to be added here.

src/data/game-score.ts
```diff
@@ -23,10 +52,23 @@ export class GameScore {
...
+          //Update if ONLY NEW SCORE IS HiGHER and mini game score is the same.
+          const savedMiniGameScore = allGameLevelInfo[index].treasureChestMiniGameScore;
+          //Used the saved value of mini game score.
+          levelPlayedInfo.treasureChestMiniGameScore = savedMiniGameScore;
+        }
+        // Save the updated score with the preserved mini game score.
         allGameLevelInfo[index] = levelPlayedInfo;
+     } else {
+      //If new game score IS NOT higher and ONLY the MINI GAME SCORE is higher. Otherwise do nothing.
+      if (isNewMiniGameScoreHigher){
```
janfb-codev, line 67: Can we move this if condition on the else at line 65?

src/data/game-score.ts
janfb-codev: This is still not typed/interface? Not sure if getAllGameLevelInfo is being used elsewhere but if we have this type maybe we can return it there as that type also instead of any[].

One of the benefits of this change if ever getAllGameLevelInfo is used in other parts of the codebase, we will be able to use intellisense to get the properties of this data without looking back and forth on how was the data structured.

src/data/game-score.ts
bernhardccodev: Hi @janfb-codev , clean up added. As for your second comment regarding `getAllGameLevelInfo `, I can't modify that as it might affect other classes unrelated to this PR.

src/data/game-score.ts
janfb-codev: This change will probably affect a lot of other classes and even flow. This is just good to have.

janfb-codev (approved):

PR #1808, opened 2025-09-15
feat: FM-656 Mini Game animation transition and optimization of minigame initialization logic

ashwinnair-chimple:
# Changes
- Mini Game animation transition 
- optimization of minigame initialization logic
- chore: added local link for dev debugging
 and improved logic
- code cleanup and added helper methods for draw loop

# How to test
- Play puzzles and check mini game transition

Ref: [FM-656](https://curiouslearning.atlassian.net/browse/FM-656)


src/gamepuzzles/feedbackAudioHandler/feedbackAudioHandler.ts
```diff
@@ -100,8 +103,14 @@ export default class FeedbackAudioHandler {
...
         )
       ]);
+      setTimeout(() => {
+        if (onFeedbackAudioEnd) onFeedbackAudioEnd(); // Callback after audios finish playing
+      }, 4500);
     } catch (error) {
       console.warn('Audio playback failed:', error);
+      setTimeout(() => {
+        if (onFeedbackAudioEnd) onFeedbackAudioEnd(); // Ensure callback is called even if audio fails
+      }, 4500);
```
bernhardccodev, line 113: Hi @ashwinnair-chimple , are these changes for the `feedbackAudioHandler `only to sync with the mini game or are all of these will reflect on every game level?

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -361,7 +361,10 @@ export class GameplayScene {
         timerTicking: this.timerTicking,
         lang,
         lettersCountRef,
-        feedBackTexts: this.feedBackTexts
+        feedBackTexts: this.feedBackTexts,
+        onFeedbackAudioEnd: (isCorrect) => {
+          this.determineNextStep(isCorrect); // <-- This will be call minigame if level is for minigame.
+        }
```
bernhardccodev, line 367: Hi @ashwinnair-chimple , I think there is an issue here. If this is only for mini game feature. The problem is that `this.determineNextStep(isCorrec)` will be called twice. One from this new call back you created and another one in  `handleStoneDropEnd` method of gameplay-scene class.

bernhardccodev (changes requested):
Hi @ashwinnair-chimple, great work on this. I do have a few concerns:

Adding more delays in feedbackAudioHandler with setTimeout makes the flow more complex, especially for syncing audio with the mini game. The new delays also seem to work only for levels without a mini game.

The callback in gameplay-scene risks invoking this.determineNextStep(isCorrect) twice — once from the feedbackAudio callback you added and again from handleStoneDropEnd.

Instead of adding nested callbacks and adjusting setTimeout, could we leverage the MiniGameStateService pub/sub system that was designed for these scenarios? That way, we don’t need to touch the audio flow (which is already working fine) — we’d just publish from the mini game and subscribe in the mini game handler.

What do you think?

src/miniGame/miniGames/treasureChest/treasureChestAnimation.ts
```diff
@@ -214,137 +217,146 @@ export class TreasureChestAnimation {
...
-        this.hide();
-        if (this.onFadeComplete) {
-          this.onFadeComplete();
-          this.onFadeComplete = undefined; // prevent double calling
-        }
+    // Fade check
+    if (this.handleFadeOut()) {
+      if (!this.fadeOutStart) {
+        // fadeOutStart was active, now finished
         return;
```
janfb-codev, line 224: Won't this cause an issue since upon reaching here it will return the whole loop and will not do the remaining of this function?

bernhardccodev (changes requested):
Hi @ashwinnair-chimple , you are almost there.

I know I originally mentioned using `miniGameStateService`, but looking at your approach I can see that I was mistaken—it really should be `gameStateService`.

I noticed you’re still using `miniGameStateService `along with some callback logic and conditional tweaks in `handleStoneDropEnd`.

What I’m suggesting instead is to rely on `gameStateService` and introduce a new event, e.g. `LOAD_THE_NEXT_LEVEL`, for pub/sub.

That way, in `gameplay-scene.ts`, instead of callbacks and extra conditions, you could simply store `isCorrect` in `this.isCorrect` and then:
```
this.unsubscribeLoadPuzzle(
  gameStateService.EVENTS.LOAD_THE_NEXT_LEVEL,
  () => {
    this.determineNextStep(this.isCorrect);
  }
);
```
This lets `determineNextStep `handle whether to show the mini game or load the puzzle.

Then in feedbackAudioHandler.ts, instead of adding delays, you’d just publish the event:
```
gameStateService.publish(
  gameStateService.EVENTS.LOAD_THE_NEXT_LEVEL,
  <any value here>
);
```

This way we avoid delays, callbacks, and special-case tweaks for mini game levels—just publish at the right time, and gameplay will take care of the rest.

Thoughts?

bernhardccodev (approved):
Hi @ashwinnair-chimple , the logic looks solid now! The only thing i can see are the unused codes in puzzle-handler and the unused event in miniGameStateService.ts.

But overall looks good, approved.

PR #1818, opened 2025-09-25
fix: [FTM-689] Added DeltaTime for Treasure Stones Animation. This fi…

janfb-codev:
…x is needed since different device has different refresh rate.

# Changes
- Added DeltaTime on TreasureStones Animation and Movement

# How to test
- You can test this on different device that has different refresh rate.
- Testing on Laptop or on Monitor, you need to change your refresh rate on different amount to test

Ref: [<JIRA-CODE>](https://curiouslearning.atlassian.net/browse/FTM-689)


bernhardccodev (approved):
Looks great!

PR #1896, opened 2026-01-22
fix: FM-818 Stones From Previous Puzzle Remain Interactive After Mini-Game Trigger on Timer End.

ashwinnair-chimple:
# Changes
- fixed Stones From Previous Puzzle Remain Interactive After Mini-Game Trigger on Timer End.
- chore: reduce delay after minigame to puzzle load to 1.5s from 4.5s 

# How to test
- Open the app and click on the play button and select a level which has the mini game.
- Start the level and let the timer run out.
- Once the mini game is triggered and completed, the flow will continue from the puzzle where the timer ran out.
- Here we can drag and drop the stones to the monster which will break the normal flow.

Ref: [<JIRA-CODE>](https://curiouslearning.atlassian.net/browse/<JIRA-CODE>)


bernhardccodev (approved):

janfb-codev (approved):
Looks good to me. Changelog.md package-lock.json, package.json should be removed after merge.

PR #1897, opened 2026-01-22
feat: Scheduler Implementation for Pause and Resume Mechanic (FM-806)

janfb-codev:
# Changes
- Implemented Scheduler for the pause mechanic (FM-807)
- Used Scheduler for Gameplay (FM-808)
- Fixed issue with the timer rotation not resuming after paused before timer (FM-817)

# How to test
- Pause and Resume at different stages of the game
- Pause when the stones are just flying thru
- Pause when the timer starts
- Pause after completing a puzzle
- Pause in Minigame via switching to different tab since pause button is not clickable at Minigame

Ref: [FM-806](https://curiouslearning.atlassian.net/browse/FM-806)

[FM-806]: https://curiouslearning.atlassian.net/browse/FM-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Enhanced pause/resume functionality for game audio and animations with improved state management.
  * Improved animation timing synchronization across game elements.

* **Bug Fixes**
  * Fixed event subscription cleanup to prevent memory leaks.
  * Improved game state consistency during pause and resume.

* **Chores**
  * Version bumped to v1.3.2.
  * Internal game architecture improvements for better maintainability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


src/services/scheduler.ts
```diff
@@ -0,0 +1,112 @@
...
+      loop: true,
+    });
+    return id;
+  }
+
+  /**
+   * Cancels a previously scheduled interval.
+   * @param id The ID of the timer to clear.
+   */
+  clearInterval(id: TimerId): void {
```
miguelccodev: Minor

Avoid aliases/methods that do the same thing.

Suggestion: rename cancellation instead of clearTimeout/clearInterval.

src/services/scheduler.ts
```diff
@@ -0,0 +1,112 @@
...
+        try {
+          timer.callback();
+        } catch (e) {
+          console.error("Error in scheduled callback:", e);
+        }
+
+        if (timer.loop) {
+          timer.remaining += timer.delay;
+        } else {
+          this.timers.delete(timer.id);
```
miguelccodev: Minor

At the moment, there is no difference in calling clearTimeout. But for future consistency, call that instead of deleting and item from the map directly.

src/services/scheduler.ts
```diff
@@ -0,0 +1,112 @@
...
+  loop: boolean;
+};
+
+let nextTimerId = 0;
+
+/**
+ * Custom scheduler for managing time-based events within the game loop.
+ * Unlike standard window.setTimeout/setInterval, this respects the game's 
+ * pause state and is driven by deltaTime from the main update loop.
+ */
```
miguelccodev, line 20: Minor

Add TODO note that this should be migrated into core module repository for easy reuse.

src/services/scheduler.ts
```diff
@@ -0,0 +1,112 @@
+import { Opaque } from 'type-fest';
```
miguelccodev, line 1: Moderate

Write minimal happy-path unit test for scheduler using gherkin syntax. You can prompt ai and modify.

miguelccodev (changes requested):
Added comments. Please have a look.

src/components/audio-player.ts
```diff
@@ -1,10 +1,13 @@
 import { Window } from "@common";
 import { AUDIO_PATH_BTN_CLICK } from "@constants";
+import scheduler from "../services/scheduler";
```
bernhardccodev: Hi @janfb-codev I think we should use aliases here for quick path access.

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -417,7 +435,7 @@ export class GameplayScene {
   }
 
   public handleVisibilityChange(): void {
-    this.audioPlayer.stopAllAudios();
+    // this.audioPlayer.stopAllAudios();
```
bernhardccodev: Lets removed unused code here.

src/common/stone-config/stone-config.spec.ts
```diff
@@ -76,19 +76,19 @@ describe('StoneConfig', () => {
       stoneConfig.initialize();
       stoneConfig['animationStartTime'] = 0;
       now = 0;
-      stoneConfig.draw();
+      stoneConfig.draw(0);
```
bernhardccodev, line 79: Hi @janfb-codev , wouldn't be simpler if we just set a default value of 0 in the draw method of stoneConfig like how the shouldResize value is default to false incase there is no param passed?

src/components/audio-player.ts
```diff
@@ -29,30 +33,32 @@ export class AudioPlayer {
...
-    if (!this.isClickSoundLoaded) {
-      // Load and decode the audio on demand if it hasn't been loaded
-      try {
-        this.clickSoundBuffer = await this.loadAndDecodeAudio(audioSrc);
-        AudioPlayer.audioBuffers.set(audioSrc, this.clickSoundBuffer);
-        this.isClickSoundLoaded = true; // Set the flag to true after loading
-      } catch (error) {
-        console.error("Error loading or decoding click sound:", error);
-        return;
+  playUIAudio(audioSrc: string, volume: number = 1, onEnded?: () => void) {
```
bernhardccodev, line 49: Hi @janfb-codev this is just a suggestion but it would be nice to have a comment here if this method is only for prompt audio or it can be used in any other parts where audio is needed.

src/components/audio-player.ts
```diff
@@ -281,11 +299,20 @@ export class AudioPlayer {
...
+  private static nonPausableAudioContext: AudioContext | null = null;
   static getAudioContext(): AudioContext {
     if (!AudioContextManager.instance) {
       AudioContextManager.instance = new (window.AudioContext ||
         (window as Window).webkitAudioContext)();
     }
     return AudioContextManager.instance;
   }
+
+  static getNonPausableAudioContext(): AudioContext {
```
bernhardccodev, line 422: Hi @janfb-codev same here, just a suggestion to specify that this should be use to the components such as popup components or any other related feature that should play an audio and not be affected by pause.

bernhardccodev (approved):
Hi @janfb-codev looks good to me. I have reviewed it carefully and I am alright with the overall implementation.

I left some comments but it is more polishing and clean up and optional. Thanks!

src/common/stone-config/stone-config.spec.ts
janfb-codev: The parameter of draw is a required value setting it to default to 0 could be missed when there is a change in the code.

miguelccodev (approved):

src/components/prompt-text/prompt-text.ts
```diff
@@ -219,7 +221,7 @@ export class PromptText extends BaseHTML {
     }
 
     private runAfterInitialAudioDelay(delayMs: number, callback: () => void) {
-        setTimeout(callback, delayMs);
+        scheduler.setTimeout(callback, delayMs);
```
ashwinnair-chimple, line 235: The scheduler.setTimeout calls in runAfterInitialAudioDelay (and handleAutoPromptPlay) don't store timer IDs. If the component is disposed while timers are pending, callbacks may execute on destroyed elements, potentially causing errors or memory leaks.

instead we can store timer IDs and clear them in dispose

src/components/prompt-text/prompt-text.ts
```diff
@@ -555,6 +557,18 @@ export class PromptText extends BaseHTML {
         }
     }
 
+    private handleGamePause(isPaused: boolean) {
```
ashwinnair-chimple, line 571: This method is defined but never called.

src/services/scheduler.ts
```diff
@@ -0,0 +1,105 @@
...
+   * @param delay Interval in milliseconds.
+   * @returns A unique TimerId for clearing the interval.
+   */
+  setInterval(callback: () => void, delay: number): TimerId {
+    const id = nextTimerId++ as TimerId;
+    this.timers.set(id, {
+      id,
+      callback,
+      delay,
+      remaining: delay,
```
ashwinnair-chimple: A zero or negative delay would cause the callback to execute on every update() tick, which could significantly impact performance. Consider adding a minimum delay floor for intervals

const safeDelay = Math.max(1, delay); // Minimum 1ms to prevent runaway intervals

src/components/audio-player.ts
```diff
@@ -29,32 +38,50 @@ export class AudioPlayer {
...
+   * Plays a UI audio effect (non-pausable).
+   * Used for interface sounds like button clicks that should persist even when the game is paused.
+   * @param audioSrc - The source path or identifier of the audio to play.
+   * @param volume - The volume level (default: 1).
+   * @param onEnded - Optional callback to execute when audio playback ends.
+   * @returns The created AudioBufferSourceNode if successful, or null.
+   */
+  playUIAudio(audioSrc: string, volume: number = 1, onEnded?: () => void) {
+    const audioBuffer: AudioBuffer = AudioPlayer.audioBuffers.get(audioSrc);
+    if (audioBuffer) {
```
ashwinnair-chimple: Also add a check for nonPausableAudioContext along with audioBuffer to prevent null
`if (audioBuffer && this.nonPausableAudioContext) `

src/components/timer-ticking.ts
```diff
@@ -50,7 +51,7 @@ export default class TimerTicking {
         this.timerHtmlComponent = new TimerHTMLComponent('timer-ticking');
         // Reference the container element for the "full timer" image
         // Verify and cache the DOM element after rendering
-        setTimeout(() => {
+        scheduler.setTimeout(() => {
```
ashwinnair-chimple, line 54: The timeout handle returned by scheduler.setTimeout is not stored, so it cannot be cancelled when the component is destroyed. If destroy() is called before this timeout fires, the callback will execute on a destroyed instance, potentially causing errors.

Store the timeout handle and cancel it in destroy()

src/gamepuzzles/feedbackAudioHandler/feedbackAudioHandler.ts
```diff
@@ -72,7 +73,7 @@ export default class FeedbackAudioHandler {
    */
   private playIncorrectFeedbackSound(): void {
 
-    setTimeout(() => {
+    scheduler.setTimeout(() => {
```
ashwinnair-chimple, line 76: The scheduler.setTimeout returns a timer ID that is discarded. If dispose() is called while this timer is pending, audioEndCallback() will still fire and publish LOAD_NEXT_GAME_PUZZLE, potentially causing state inconsistencies or errors.

Consider storing timer IDs as class members and clearing them in dispose(

src/miniGame/miniGames/treasureChest/treasureChestMiniGame.spec.ts
```diff
@@ -38,7 +38,7 @@ describe('Testing TreasureChestMiniGame.', () => {
     miniGame['collectedBeforeThreshold'] = 3;
 
     // Run draw to trigger animation
```
ashwinnair-chimple: The comment still references "draw" but the code now calls start().

src/scenes/gameplay-scene/monster-controller.ts
```diff
@@ -84,11 +85,19 @@ export class MonsterController {
   public triggerMonsterAnimation(animationName: string): void {
     const delay = this.animationDelays[this.monsterPhaseNumber]?.[animationName] ?? 0;
     if (delay > 0) {
-      setTimeout(() => this.monster?.triggerInput(animationName), delay);
+      scheduler.setTimeout(() => this.monster?.triggerInput(animationName), delay);
```
ashwinnair-chimple, line 88: Same point to store and cancel on dispose

src/miniGame/miniGames/treasureChest/treasureChest.ts
```diff
@@ -44,14 +41,22 @@ export default class TreasureChest {
     const chestY = height - chestH - 20;
 
     // Pulse/rotate chest if just spawned (< 1s old)
-    const elapsed = time - this.lastSpawnTime;
+    // For pulse, we use the time since state started (which is passed as time, if stateTimer is passed)
```
ashwinnair-chimple: Remove stale code

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -94,6 +95,8 @@ export class GameplayScene {
   isFeedBackTriggered: boolean;
   // #endregion
 
+  private isPaused: boolean;
```
ashwinnair-chimple: Initialize isPaused to avoid relying on undefined coercion.

private isPaused: boolean = false;

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -319,14 +333,18 @@ export class GameplayScene {
   }
 
   public resumeGame(): void {
+    this.isPaused = false;
+    this.audioPlayer.resumeAllAudios();
     // Resume the clock rotation when game is resumed
-    this.uiManager.applyTimerRotation(this.uiManager.timerTicking?.startMyTimer && !this.uiManager.timerTicking?.isStoneDropped);
+    this.monsterController.resume();
```
ashwinnair-chimple: add check this.monsterController?.resume();

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -319,14 +333,18 @@ export class GameplayScene {
...
   }
 
   public pauseGamePlay(): void {
-    this.audioPlayer.stopAllAudios();
+    this.isPaused = true;
+    this.audioPlayer.pauseAllAudios();
```
ashwinnair-chimple: addc check  this.audioPlayer?.pauseAllAudios(); and this.monsterController?.pause();

src/components/prompt-text/prompt-text.ts
```diff
@@ -195,6 +198,16 @@ export class PromptText extends BaseHTML {
                 }
             )
         );
+
+        //Subscription for game starting.
```
ashwinnair-chimple: change comment to game pause

src/components/timer-ticking.ts
ashwinnair-chimple: this need to be handled @janfb-codev

src/miniGame/miniGames/treasureChest/treasureChest.ts
ashwinnair-chimple: @janfb-codev remove this notes

src/miniGame/miniGames/treasureChest/treasureChestAnimation.ts
```diff
@@ -314,24 +290,24 @@ export class TreasureChestAnimation {
...
         const alpha = Math.max(0, 1 - elapsed / this.fadeOutDuration);
         this.ctx.globalAlpha = alpha;
         this.treasureChest.drawOpenChest(this.width, this.height);
         this.treasureStone.stoneBurstAnimation(this.width, this.height, deltaTime);
 
         //Keep showing Blue Star if still active
-        this.drawBlueBonusStar(time);
+        this.drawBlueBonusStar(deltaTime);
 
         if (alpha === 0) {
```
ashwinnair-chimple: Exact float comparison may cause fade-out to never complete.
alpha === 0 relies on exact floating-point equality. Due to floating-point precision, alpha might be a very small negative number (e.g., -0.0001) rather than exactly 0, causing the fade-out to never complete.

ashwinnair-chimple (commented):
@janfb-codev I have added and updated few comments can you resolve that?

Thanks!

src/components/timer-ticking.ts
janfb-codev: The scheduler itself records the timers id and handle it inside the Scheduler.

src/scenes/gameplay-scene/monster-controller.ts
janfb-codev: The scheduler itself records the timers id and handle it inside the Scheduler as well.

src/components/audio-player.ts
```diff
@@ -246,6 +301,29 @@ export class AudioPlayer {
     this.audioSourcs = [];
   };
 
+  /**
```
ashwinnair-chimple, line 333: suspend() and resume() return Promises that are not handled.

AudioContext.suspend() and AudioContext.resume() are asynchronous and return Promises. The current implementation doesn't await them or handle potential rejections.

src/components/audio-player.ts
```diff
@@ -29,32 +38,50 @@ export class AudioPlayer {
...
+    const audioBuffer: AudioBuffer = AudioPlayer.audioBuffers.get(audioSrc);
+    if (audioBuffer && this.nonPausableAudioContext) {
+      const sourceNode = this.nonPausableAudioContext.createBufferSource();
+      const gainNode = this.nonPausableAudioContext.createGain();
+      sourceNode.buffer = audioBuffer;
+      sourceNode.connect(gainNode);
+      gainNode.connect(this.nonPausableAudioContext.destination);
+      // handle end of playback
+      if (onEnded) {
+        sourceNode.onended = onEnded;
```
ashwinnair-chimple: Source nodes are pushed to audioSourcs (line 64) but never removed when playback ends. For UI sounds played frequently (e.g., button clicks), this array grows unboundedly, leaking memory.

sourceNode.onended = () => {
          this.removeSourceNode(sourceNode);
          onEnded();
        };
      } else {
       sourceNode.onended = () => this.removeSourceNode(sourceNode);
       }

Add a helper method to the class:

private removeSourceNode(node: AudioBufferSourceNode): void {
  const index = this.audioSourcs.indexOf(node);
  if (index > -1) {
    this.audioSourcs.splice(index, 1);
  }
}

src/scenes/gameplay-scene/gameplay-scene.ts
```diff
@@ -319,14 +333,18 @@ export class GameplayScene {
   }
 
   public resumeGame(): void {
+    this.isPaused = false;
+    this.audioPlayer?.resumeAllAudios();
     // Resume the clock rotation when game is resumed
```
ashwinnair-chimple: // Resume the timer rotation and monster actions when game is resumed
  this.uiManager.applyTimerRotation(true);

janfb-codev: Regarding this comment, the timer rotation resumes on the update part of the TimerTicking. Resuming the timer here will resume the rotation animation even though the timer hasn't started.

ashwinnair-chimple: Got it ! As there was comment on the same i thought it was missing on resume.

ashwinnair-chimple (approved):
all changes applied and looks good now!
Thanks @janfb-codev

PR #1900, opened 2026-01-29
Develop to test code merge

ashwinnair-chimple:
Fixed the issue In the gameplay after we complete level 1, from level 2 we can observe the stones are appearing even before the target audio plays.


janfb-codev (approved):
Looks good to me!

bernhardccodev (approved):

PR #2013, opened 2026-08-17
fix: [MR-177] Added logInitialSummaryData and schema of summary data

janfb-codev:
# Changes
- FTM now seeds its Firestore summary document once per launch, writing `0` for every numeric
  summary field. Previously a field only appeared after its first matching event, so a player who
  quit mid-level had no `levels_completed`, `highest_level_completed` or `time_spent_total_second`
  key at all, which read as nulls downstream.
- Added `src/modules/android/constants/summary-data-schema.ts` as the single source of truth for the
  summary schema: the `FtmSummaryData` interface plus `FTM_SUMMARY_DATA_DEFAULTS` typed
  `Required<FtmSummaryData>`, so adding a field to the interface fails the build until it is given a
  default. Adding a field also re-seeds existing documents on the next launch.
- `AndroidInterface` is now constructed as `AndroidInterface<FtmSummaryData>`, so every summary write
  and its per-field options are checked against the schema at compile time, an undeclared field, or
  an options key naming one, no longer compiles.
- Bumped `@curiouslearning/core` to `^1.13.1`, which adds `logInitialSummaryData()`, `isAvailable()`
  and the once-per-document seed guard.

Seeding is safe to repeat: core sends every field with the `"add"` instruction, which the container
maps to `FieldValue.increment`, so a `0` creates an absent field and is a no-op on a real value.

# How to test
- Manual, in Curious Reader (requires `mr-75` enabled for the test user, or the Android strategy is
  never registered and nothing is sent):
  - Launch FTM with a fresh `cr_user_id` and **complete nothing**. A `summary_data` document should
    appear with all six fields at `0`, and `created_at == updated_at`.
  - Complete one puzzle. `puzzles_completed` and `puzzle_success` go `0 → 1` in the *same* document,
    with no second document created.
  - Relaunch. No new write — `updated_at` unchanged.
  - Switch language and launch. A separate document is seeded for that language.
  - Open FTM outside the container. Nothing is sent, and a later in-container launch still seeds.

Ref: [MR-177](https://curiouslearning.atlassian.net/browse/MR-177)

[MR-177]: https://curiouslearning.atlassian.net/browse/MR-177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@capacitor/core": "^4.4.0",
         "@curiouslearning/analytics": "^1.3.1",
         "@curiouslearning/assessment-survey": "^1.2.1",
-        "@curiouslearning/core": "^1.13.0",
+        "@curiouslearning/core": "^1.13.1",
         "@curiouslearning/features": "^1.3.0",
         "@octokit/rest": "^20.0.1",
         "@release-it/conventional-changelog": "^8.0.1",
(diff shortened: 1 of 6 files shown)


bernhardccodev (approved):

ashwinnair-chimple (approved):
approved

